PR #1, opened 2022-09-16
document API, extend tests a little

mixmix:
@Powersource @staltz 

I did some "documentation driven development" and surfaced some new needs/ requirements.
My recommendations here are based on learning on `ssb-tribes` and how I wish that API was, and also where I know parts of this might need to be able to go

Note that group profiles are a seperate concern in `ssb-tribes`. I think we should keep that the same and leave them out of here for the moment


test/testbot.js
```diff
@@ -0,0 +1,17 @@
+const Bot = require('scuttle-testbot')
```
staltz, line 1: Gotta fix the license header :)

test/testbot.js
```diff
@@ -0,0 +1,17 @@
...
+
+module.exports = function Testbot (opts = {}) {
+  const stack = Bot
+    // NOTE: these already already in scuttlebot
+    // .use(require('ssb-db2'))
+    // .use(require('ssb-box2'))
+    .use(require('../'))
```
staltz, line 8: @mixmix I know you like this tool, but I don't think it's helping us in this case. I doesn't include ssb-box2, as far as I can see.

staltz: Also, we can't get to choose the version of ssb-db2 (e.g. version 6).

README.md
```diff
@@ -19,19 +19,69 @@ npm install ssb-tribes2
...
+
+ssb.tribes2.start()
+
+ssb.tribes2.create({}, (err, group) => {
+  const content = {
+    type: 'post',
+    text: 'welcome to the group',
+    recps: [group.id]
+  }
+  ssb.db.publishAs(group.subfeed, content, cb)
```
staltz, line 43: We should use `ssb.db.create` not `ssb.db.publishAs`

README.md
```diff
@@ -19,19 +19,69 @@ npm install ssb-tribes2
...
 ```
 
-TODO TODO
+## API
+
+
+### `ssb.tribes2.create(opts, cb)`
+
+Creates a new private group.
+This creates an encruption key, sets up a sub-feed for the group, and initializes the
```
staltz, line 53:
```suggestion
This creates an encryption key, sets up a sub-feed for the group, and initializes the
```

README.md
```diff
@@ -19,19 +19,69 @@ npm install ssb-tribes2
...
+group with a `group/init` message, and `group/add-member` to signal you were added.
+Calls back with important info about the group
+
+- `opts` *Object* - currently empty, but will be used to specifiy details like whether the group has an admin subgroup etc. in future
```
staltz, line 57:
```suggestion
- `opts` *Object* - currently empty, but will be used in the future to specify details like whether the group has an admin subgroup, etc
```

README.md
```diff
@@ -19,19 +19,69 @@ npm install ssb-tribes2
...
+- `cb` *Function* - callback function of signature `(err, group)` where `group` in object containing:
```
staltz, line 58:
```suggestion
- `cb` *Function* - callback function of signature `(err, group)` where `group` is an object containing:
```

README.md
```diff
@@ -19,19 +19,69 @@ npm install ssb-tribes2
...
+    - `id` *CloakedId* - a cipherlink that's safe to use publicly to name the group, and is used in `recps` to trigger encrypting messages to that group, encoded as an ssb-uri
+    - `subfeed` *Keys* - the keys of the subfeed you should publish group data to
+    - `secret` *Buffer*  - the symmetric key used for encryption by the group
```
staltz, line 61:
```suggestion
    - `secret` *Buffer* - the symmetric key used by the group for encryption
```

staltz (changes requested):

README.md
staltz:
```suggestion
  const keys = group.mySubfeedKeys
  const recps = [group.id]
  const content = { type: 'post', text: 'welcome, friends!' }
  ssb.db.create({ keys, recps, content }, cb)
```

staltz (changes requested):

test/testbot.js
mixmix: Why is it needed per file?

staltz: The code we're producing in this grant is under the conditions that it's open source, and to guarantee that our compliance with licenses is clear and legally correct, we should use https://reuse.software/ by the Free Software Foundation Europe that automatically checks that all source files in a project are correctly self-describing what their legal status is. The CI job checks that for us and turns red when there is a problem. (PS: let's aim for CI to always be green, even if you're just adding templates/drafts, the tests should pass).

test/testbot.js
mixmix: Yeah I'm fine too pull it out and copy in the functionality I want from that module. I like the pattern of having the tesbot util. I have some strong opinions about writing ssb tests

Namely - start and stop a server for each test. The main reason is keeping the database clean and clear makes running  standalone  tests easy with test.only, and you never accidentally make graphs of coupled tests by leaving a mess of messages which effect many tests

test/testbot.js
mixmix: Oh good. I didn't know about that, that's clear, thanks

test/testbot.js
staltz: Agree with the second paragraph!

README.md
mixmix: What's the difference? I guess I don't fully understand why we need create - apart from being about to explicitly tell it to use box2 encryption?

staltz: Oh yeah, good point, we also need to add `encryptionFormat: 'box2'` to the opts.

The difference is that publishAs just calls `create` under the hood, and we'd (arj and I) would like to deprecate publish and publishAs in the near future. create() can do everything we need, and it can be misleading to use publish or publishAs because they don't properly support choosing feed formats and encryption formats.

package.json
```diff
@@ -28,8 +28,14 @@
     "*.js"
   ],
   "dependencies": {
+    "envelope-js": "^1.3.2",
+    "envelope-spec": "^1.1.0",
+    "private-group-spec": "^1.1.3",
     "pull-stream": "^3.6.14",
-    "ssb-meta-feeds": "~0.30.0"
+    "ssb-bfe": "^3.6.0",
+    "ssb-crut": "^4.6.0",
```
staltz, line 36: @Powersource Perhaps an easy task for you would be to update ssb-crut's license from AGPL-3.0 to LGPL-3.0-only? It's something we all already agreed on. https://gitlab.com/ahau/lib/ssb-crut

index.js
```diff
@@ -31,7 +37,65 @@ module.exports = {
...
+        type: 'group/init',
+        tangles: {
+          group: { root: null, previous: null },
+        },
+      }
+      //if (!initSpec.isValid(content)) return cb(new Error(initSpec.isValid.errorsString))
+
+      /* enveloping */
+      // we have to do it manually this one time, because the auto-boxing checks for a known groupId
+      // but the groupId is derived from the messageId of this message (which does not exist yet
```
staltz, line 51: I can kind of see what's going on here (would appreciate more context, though!), but using getFeedState as a shortcut to build a ciphertext is bypassing all the work we did with `ssb.db.create`, and building some new technical debt. Let's not

lib/build-group-id.js
```diff
@@ -0,0 +1,59 @@
...
+const LABELS = require('envelope-spec/derive_secret/constants.json')
+const { keySchemes } = require('private-group-spec')
+
+const bfe = require('ssb-bfe')
+
+module.exports = function buildGroupId ({ groupInitMsg, readKey, msgKey, groupKey }) {
+  const msgId = bfe.encode(groupInitMsg.key)
+
+  if (!readKey) {
+    if (groupInitMsg.value.meta && groupInitMsg.value.meta.unbox) {
```
staltz, line 13: This function was copy pasted from ssb-tribes, but it doesn't work. ssb-db2 doesn't have `msg.value.meta.unbox`.

index.js
staltz: I took a look at the modules, and I suggest we update `ssb-box2` so that it can receive, besides strings, also `{key: groupKey.toBuffer(), scheme: keySchemes.private_group}`.

package.json
Powersource: Done, waiting on (i guess mix?) merge and new version published

mixmix: Need to publish

index.js
mixmix: We need to access the previous msgId for the feed we're publishing on.
I used this as a quick placeholder to get Jacob underway. Was going to talk with you and read docs to figure out how to get it without compat.

Basically the init message is special and we have to do manual encryption for first message. Ohhhh unless we can put keys in which are { key, scheme } yes...

test/testbot.js
Powersource: added

test/testbot.js
Powersource: @staltz you mean avoiding `scuttle-testbot` entirely? What's the alternative?

staltz: @Powersource Yeah perhaps you can do like this PR did: https://github.com/ssbc/ssb-meta-feeds/pull/73

package.json
staltz: This can be done after merging this PR.

README.md
Powersource: @staltz done

staltz: @Powersource There are two `const content` definitions still in that readme

test/testbot.js
Powersource: @staltz done

staltz: I did this final step

README.md
staltz: I did this final step

index.js
staltz: Pending on https://github.com/ssbc/ssb-box2/pull/17 merged first

Powersource: @staltz fixed

index.js
```diff
@@ -11,33 +14,93 @@ const {
...
+          if (err) return cb(err)
+
+          const data = {
+            id: buildGroupId({ groupInitMsg, msgKey }),
+            secret: groupKey.toBuffer(),
+            root: groupInitMsg.key,
+            subfeed: '',
+          }
+
+          ssb.box2.addGroupKey(data.id, data.secret)
```
staltz, line 70: We need to create a new version of ssb-box2 and update this to `addGroupInfo`

staltz (approved):
Looks good to merge, minus fixing tests

PR #2, opened 2022-09-21
Add addMembers

Powersource:
based on https://github.com/ssbc/ssb-tribes2/pull/1

### DIFF
--- a/index.js
+++ b/index.js
@@ -64,8 +64,7 @@ module.exports = {
             subfeed: '',
           }
 
-          // TODO: add root on new version of box2
-          ssb.box2.addGroupKey(data.id, data.secret)
+          ssb.box2.addGroupInfo(data.id, { key: data.secret, root: data.root })
 
           // TODO later: add myself for recovery reasons
 
@@ -83,7 +82,7 @@ module.exports = {
         cb(null, {
           id,
           secret: info.key,
-          //TODO: add root
+          root: info.root,
         })
       })
     }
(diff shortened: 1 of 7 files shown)


index.js
```diff
@@ -97,9 +96,44 @@ module.exports = {
       )
     }
 
-    function addMembers(groupId, feedIds, cb) {
+    function addMembers(groupId, feedIds, opts = {}, cb) {
+      if (cb === undefined) return promisify(addMembers)(groupId, feedIds, opts)
+
       // TODO
       // copy a lot from ssb-tribes but don't use the keystore from there
+      const { key, root } = get(groupId)
```
staltz, line 104: I think `get` is async and needs a cb

Powersource: ah lol thx

index.js
```diff
@@ -101,37 +102,42 @@ module.exports = {
 
       // TODO
```
staltz, line 103: Just as a sanity check, you can add a check for `feedIds.length === 0` and call `cb(new Error` in case it's true. Also, you can be more strict about it and check if `feedIds` is an array AND with positive length AND with length less than `$MAX_PERMITTED_ACCORDING_TO_SPEC` AND that every item in the array is an ssb-ref or ssb URI.

Powersource: added some simple sanity checking

Powersource:
oops licenses

Powersource:
added license headers now

staltz (approved):
Looks good. I'm a bit bothered by the commented code, but let's aim to remove all of those (including TODOs) once things have settled.

Powersource:
Yeah, slow and steady :)

PR #5, opened 2022-10-07
Add listMembers method

Powersource:
Based on https://github.com/ssbc/ssb-tribes2/pull/4

PR #8, opened 2022-10-09
Enable tanglePrune

Powersource:
Based on https://github.com/ssbc/ssb-tribes2/pull/5


Powersource:
A few issues for the future popped up
https://github.com/ssbc/ssb-tribes2/issues/9
https://github.com/ssbc/ssb-tribes2/issues/10 (this is maybe the next one i'll do)
https://github.com/ssbc/ssb-tribes2/issues/11

Also found a bug (fixed it here) that also exists in ssb-tribes https://github.com/ssbc/ssb-tribes/issues/70

test/tangle-prune.test.js
```diff
@@ -0,0 +1,122 @@
...
+const randomChar = () => chars.sort(() => (Math.random() < 0.5 ? -1 : +1))[0]
+const randomText = (length) => {
+  let output = ''
+  while (output.length < length) output += randomChar()
+  return output
+}
+
+test('tangle prune', async (t) => {
+  const ssb = Testbot()
+  const ssbId = ssb.id
```
mixmix, line 21: :hot_pepper: (emoji for potential spicy problem)
With different feedFormats, the "id" of an feedId is of different lengths, this means that if `recps` is a GroupId + 15 feedIds.... they could be "sigil" style or "ssb-uri" style.

We need to clarify in a spec "when you are DM'ing a person to add them to a group, which feedId do you use?"
Is it their rootFeedId (which is bendy butt ssb-uri?)

cc @arj03

mixmix: EASY solution @Powersource - assume all ssb-uri of the longest length, which would currently be:


https://github.com/ssbc/ssb-uri2/blob/main/test/fixtures.js#L17-L26
this long:
```
ssb:feed/gabbygrove-v1/FY5OG311W4j_KPh8H9B2MZt4WSziy_p-ABkKERJdujQ=
```
or wait... what is this
```
ssb:feed/buttwoo-v1/FY5OG311W4j_KPh8H9B2MZt4WSziy_p-ABkKERJdujQ=/Z0rMVMDEO1Aj0uPl0_J2NlhFB2bbFLIHlty_YuqArFq=
```

Yeah we need a spec

test/tangle-prune.test.js
```diff
@@ -0,0 +1,122 @@
...
+      })
+    })
+  }
+
+  async function findMaxSize(numberRecps = 1) {
+    // Apply bisection method to find max size of a message which can be published
+
+    let lower = 4000
+    let mid
+    let upper = 8000
```
mixmix, line 50: :fire: set upper to 17000

https://github.com/ssbc/ssb-buttwoo-spec#validation
> the content length in bytes. This number must not exceed 16384.

test/tangle-prune.test.js
```diff
@@ -0,0 +1,122 @@
...
+  }
+  const max16recps = await findMaxSize(16).catch(t.error) // 5546
+  const max1recp = await findMaxSize(1).catch(t.error) // 6041
+  ssb.close()
+
+  const msgId = '%RDORgMCjmL6vs51nR4bn0LWNe6wkBfbRJulSdOJsmwg=.sha256'
+  const content = (prevCount, numRecps) => ({
+    type: 'post',
+    text: 'hello!',
+    recps: new Array(numRecps).fill(ssbId),
```
mixmix, line 98: technically, this need to follow the format

```js
recps: [GroupId, FeedId, ... ]  // you need to consider the length again here 
```

mixmix (changes requested):
We need to confirm the the recps we're gonna support. See :hot_pepper:

test/tangle-prune.test.js
Powersource: Can they be URI style though? Have we allowed that anywhere actually?

test/tangle-prune.test.js
Powersource: If we change this to 17000 now then later when I do https://github.com/ssbc/ssb-tribes2/issues/10 this test won't start failing. I sort of want it to fail so I know that I actually managed to swap the format. But maybe there are more solid ways to test which format we're using?

test/tangle-prune.test.js
```diff
@@ -0,0 +1,122 @@
...
+  t.true(
+    encodedLength(result16) <= max16recps,
+    `pruned ${4000 - result16.tangles.group.previous.length}`
+  )
+
+  const result1 = tanglePrune(content(4000, 1))
+  t.true(
+    encodedLength(result1) <= max1recp,
+    `pruned ${4000 - result1.tangles.group.previous.length}`
+  )
```
Powersource, line 119: I have a feeling it's a good idea to check we're in the right ballpark and not just under a certain number. So changing this to something like this instead

```js
const e = encodedLength(result16)
t.true(e > max16recps - 200 && e <= max16recps)
```

test/tangle-prune.test.js
Powersource: Also worth considering this

https://github.com/ssbc/ssb-tribes2/blob/dc4e7bf6fbe0ececd780d7e33cdb95659d7b9577/lib/tangle-prune.js#L8-L15

staltz:
In a video call with Jacob, I suggested a different way of checking whether the content is too large, *without* hardcoding the choice of feedFormat: we can just catch the error emitted from ssb.db.create, and if that error matches "too large" (either matching the err.message string, or we can come up with error codes like err.code === 'ETOOLARGE'), then prune and retry ssb.db.create.

mixmix:
yeah that could work, could make it less resource intensive by just cutting the number of `previous` in half, and keep doing that (stopping at 1) till it works.

The tricky case to test is whether it copes when you call `create` 5000 times in a row (fast), which is what happens when a person imports a whakapapa from CSV in Ahau

Powersource:
i feel stuck, which of the options should i go with?

mixmix:
agree. Try Staltz's suggestion, and we see how bad it is!

Powersource:
fixed with https://github.com/ssbc/ssb-tribes2/pull/19 instead

Issue #10, opened 2022-10-12
Actually run it on buttwoo


Powersource:
related https://github.com/ssbc/ssb-tribes2/issues/9

right now i'm adapting this pr https://github.com/ssbc/ssb-tribes2/pull/8 to assume we're publishing on buttwoo instead of classic (like old tribes1). i've swapped json.stringify for bipf size checking but the size is still limited to 8k. buttwoo supposedly has a limit of 16k which changes things.

Powersource:
sidequest https://github.com/ssbc/ssb-keys/pull/101

staltz:
We decided to postpone using buttwoo until we can properly address (perhaps in a dedicated grant project) how to support per-message deletion in buttwoo, and we need resources to harden (fix bugs, polish, test and analyse) buttwoo for use in production.

Issue #9, opened 2022-10-11
Make stuff work on feed encodings other than buttwoo


Powersource:
Working on https://github.com/ssbc/ssb-tribes2/pull/8 there's code in `tanglePrune` that assumed how the content of a msg is stringified. In tribes1 it assumed classic with json.stringify, now I'm thinking of assuming buttwoo with bipf. But it should probably be generalized or at least documented that it (or at least that feature) only supports 1 format.

staltz:
Shouldn't be an issue any more.

PR #12, opened 2022-10-13
Run on buttwoo instead of classic

Powersource:
Based on https://github.com/ssbc/ssb-tribes2/pull/8

Fixes https://github.com/ssbc/ssb-tribes2/issues/10

Doing this for now since I think it'll cover our usecases, but in the future I guess we should do https://github.com/ssbc/ssb-tribes2/issues/9

### DIFF
--- a/index.js
+++ b/index.js
@@ -57,6 +57,8 @@ module.exports = {
           content,
           recps: recipientKeys,
           encryptionFormat: 'box2',
+          feedFormat: 'buttwoo-v1',
+          tag: 0,
         },
         (err, groupInitMsg) => {
           if (err) return cb(err)
@@ -163,11 +165,19 @@ module.exports = {
       addGroupTangle(content, (err, content) => {
         if (err) return cb(err)
 
-        ssb.db.create({ content, encryptionFormat: 'box2' }, (err, msg) => {
-          if (err) return cb(err)
+        ssb.db.create(
+          {
+            content,
+            encryptionFormat: 'box2',
+            feedFormat: 'buttwoo-v1',
+            //tag: 0,
+          },
+          (err, msg) => {
+            if (err) return cb(err)
 
-          cb(null, msg)
-        })
+            cb(null, msg)
+          }
+        )
       })
     }
 
--- a/index.js
+++ b/index.js
@@ -173,7 +173,7 @@ module.exports = {
             content,
             encryptionFormat: 'box2',
             feedFormat: 'buttwoo-v1',
-            //tag: 0,
+            tag: 0,
           },
           (err, msg) => {
             if (err) return cb(err)
--- a/index.js
+++ b/index.js
@@ -59,7 +59,9 @@ module.exports = {
           recps: recipientKeys,
           encryptionFormat: 'box2',
           feedFormat: 'buttwoo-v1',
-          tag: butt2.tags.SSB_FEED,
+          //TODO: SSB_FEED errors
+          // https://github.com/ssbc/ssb-buttwoo/issues/5
+          tag: 0, //butt2.tags.SSB_FEED,
         },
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -49,9 +49,11 @@
     "prettier": "^2.6.2",
     "pretty-quick": "^3.1.3",
     "secret-stack": "^6.4.1",
+    "ssb-buttwoo": "^0.3.2",
     "ssb-caps": "^1.1.0",
     "ssb-db2": "^6.2.3",
     "ssb-ebt": "^9.1.2",
+    "ssb-keys": "Powersource/ssb-keys#load-or-create-sync-buttwoo",
```
Powersource, line 56: https://github.com/ssbc/ssb-keys/pull/101

Powersource:
https://github.com/ssbc/ssb-uri2/issues/14

Powersource:
https://github.com/ssbc/ssb-uri2/pull/15

test/add-member.test.js
```diff
@@ -10,8 +10,8 @@ const Testbot = require('./helpers/testbot')
 const replicate = require('./helpers/replicate')
 
 test('get added to a group', async (t) => {
-  const alice = Testbot({ keys: ssbKeys.generate(null, 'alice') })
-  const bob = Testbot({ keys: ssbKeys.generate(null, 'bob') })
+  const alice = Testbot({ keys: ssbKeys.generate(null, 'alice', 'buttwoo-v1') })
+  const bob = Testbot({ keys: ssbKeys.generate(null, 'bob', 'buttwoo-v1') })
```
staltz, line 14: This is going to get quite complicated soon when we add ssb-meta-feeds to ssb-tribes2 and then it'll try to declare the buttwoo feed as the "main". Also, the fact this works now with shs is coincidence, and there may be a bunch of other places where this could break because the `@` sigil feeds are conflated with "network identity" in a bunch of places.

I recommend reverting this change, and also we don't need loadOrCreate to support other feed formats. We're going to live with classic sigil IDs as the "network identity" until the root metafeed is the primary identity, and until we start supporting https://github.com/ssbc/ssb-network-identity-spec .

index.js
```diff
@@ -163,11 +165,19 @@ module.exports = {
       addGroupTangle(content, (err, content) => {
         if (err) return cb(err)
 
-        ssb.db.create({ content, encryptionFormat: 'box2' }, (err, msg) => {
-          if (err) return cb(err)
+        ssb.db.create(
+          {
+            content,
+            encryptionFormat: 'box2',
+            feedFormat: 'buttwoo-v1',
```
staltz, line 172: If we don't use `buttwoo-v1` in loadOrCreate, then I can see that this line wouldn't work. In fact, what we need here is metafeeds, so we can create a subfeed (under a shard) which has the buttwoo feedFormat, and then pass the keys to that buttwoo identity to ssb.db.create().

I also have a meta question here, which is: why buttwoo for groups?

test/add-member.test.js
Powersource: Yes this is more of a cheap and simple way to simulate metafeeds

index.js
Powersource: my understanding was that buttwoo would be used for new feeds and that groups would only be on new feeds. but i guess we'll discuss this in an upcoming meeting.

this line does feel a bit weird. why do i have to input the feed format into create? couldn't create detect what feed format is used on the feed i'm posting to and simply use that format? or are feeds supposed to be able to contain multiple formats?

Powersource:
Latest progress

```
get added to a group
    ✖ 1) bob is a member of a group now
      Expected 1 but got 0
```

I'll put this PR on pause now until we've decided what to do with the feed formats

PR #16, opened 2022-10-14
improve replicate helper

staltz:


Powersource:
We added a timeout now anyway. Shorter timeout than before this PR so an improvement, but still weird.

Powersource:
@staltz @mixmix do you want to figure out why we need the timeout even if we check the clocks? i have no idea

staltz:
I did a bit of digging, and at the point when the clocks are synchronized, ssb-ebt has already called `ssb.db.add` and `ssb.db.add` has already answered that appending was complete, so the problem is not that the messages aren't there. The messages are definitely there as soon as clocks are synchronized.

The reason why `bob.tribes2.list()` doesn't list all the groups up-to-date yet is that there is another asynchronous task going on:

https://github.com/ssbc/ssb-tribes2/blob/c4dcfd1c4f9a78009f1cb6b6b61340349cb59d10/index.js#L191-L217

And we're not explicitly waiting for that job to be done, when we call `list()`

https://github.com/ssbc/ssb-tribes2/blob/c4dcfd1c4f9a78009f1cb6b6b61340349cb59d10/index.js#L101

TL;DR we should wait. We need to change `ssb.tribes2.list()` so that it does an ssb-db2 query for old messages in the log that have not yet been processed, and add them to box2, *before* listing everything from box2.

staltz:
So, in fact, a path forward, is to remove the setTimeout in this helper, and merge this PR. Tests will break, but that's because the implementation of `list()` is broken. Then you can work on fixing that one.

Powersource:
@staltz but won't we run into the problem of potentially unprocessed messages whenever we interact with box2? like in this function which is called whenever you publish a message to a group https://github.com/ssbc/ssb-tribes2/blob/c4dcfd1c4f9a78009f1cb6b6b61340349cb59d10/lib/get-group-tangle.js#L23

Powersource:
@staltz review/merge pls

Powersource:
if we want to look through all messages before trusting box2, why do we not just skip using box2 in that case?

staltz:
> but won't we run into the problem of potentially unprocessed messages whenever we interact with box2? like in this function which is called whenever you publish a message to a group

I guess I didn't understand this question. I was talking about `list()`, how is this related to `tribes.publish()` ?

> if we want to look through all messages before trusting box2, why do we not just skip using box2 in that case?

I also didn't understand this. I'm not that bright on understanding today.

Powersource:
Because we also need to look through all groups we've been added to when we call getGroupKeyInfo. Should we just do db2 queries instead of calling box2?

Powersource:
Could another solution be: at the end of the replicate function, wait for indexing to be done?

staltz:
> Because we also need to look through all groups we've been added to when we call getGroupKeyInfo. Should we just do db2 queries instead of calling box2?

Ah, now I understand what you mean. Yeah this is important. We do have to wait for db2 data to be up-to-date. Maybe it's better to do db2 queries instead of box2. Make it work, make it right, make it fast.

> Could another solution be: at the end of the replicate function, wait for indexing to be done?

Not quite, because there are two async tasks going on: indexing and box2.getGroupKeyInfo which does a leveldb read and this can be async with the filesystem, so it's not enough to just wait for indexing.


I'll merge and we can continue onto making the tests pass with a new implementation.

Powersource:
new pr https://github.com/ssbc/ssb-tribes2/pull/20

PR #19, opened 2022-10-21
Detect too big group tangle previous array by failing to publish, and prune it

Powersource:
As an alternative to calculating the size https://github.com/ssbc/ssb-tribes2/pull/8

### DIFF
--- a/lib/prune-publish.js
+++ b/lib/prune-publish.js
@@ -1,0 +1,22 @@
+// SPDX-FileCopyrightText: 2022 Andre 'Staltz' Medeiros <contact@staltz.com>
+//
+// SPDX-License-Identifier: LGPL-3.0-only
+
+module.exports = function prunePublish(ssb, content, cb) {
+  const group = content.tangles.group
+
+  ssb.db.create({ content, encryptionFormat: 'box2' }, (err, msg) => {
+    if (err) {
+      if (group.previous.length > 1) {
+        const halfLength = Math.ceil(group.previous.length / 2)
+        group.previous = group.previous.slice(0, halfLength)
+
+        return prunePublish(ssb, content, cb)
+      } else {
+        return cb(err)
+      }
+    }
+
+    cb(null, msg)
+  })
+}
--- a/test/tangle-prune.test.js
+++ b/test/tangle-prune.test.js
@@ -1,0 +1,60 @@
+// SPDX-FileCopyrightText: 2022 Mix Irving
+//
+// SPDX-License-Identifier: LGPL-3.0-only
+
+const test = require('tape')
+//const bipf = require('bipf')
+const Testbot = require('./helpers/testbot')
+const prunePublish = require('../lib/prune-publish')
+
+//const chars = 'abcABC123=+? '.split('')
+//const encodedLength = (obj) => bipf.encodingLength(obj)
+//const randomChar = () => chars.sort(() => (Math.random() < 0.5 ? -1 : +1))[0]
+//const randomText = (length) => {
+//  let output = ''
+//  while (output.length < length) output += randomChar()
+//  return output
+//}
+
+test('prune a message with way too big `previous`', (t) => {
+  const ssb = Testbot()
+  const ssbId = ssb.id
+
+  ssb.tribes2.create(null, (err, group) => {
+    t.error(err, 'no err on group create')
+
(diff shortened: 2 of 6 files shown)


staltz:
So far looking good!

test/tangle-prune.test.js
```diff
@@ -0,0 +1,87 @@
...
+        group: {
+          root: msgId,
+          previous: new Array(prevCount).fill(msgId),
+        },
+      },
+    })
+
+    //console.time('prune')
+    prunePublish(ssb, content(4000, 16), (err, encMsg16) => {
+      //console.timeEnd('prune')
```
Powersource, line 41: approx 3-6 ms. pruned from 4000 to 63 entries

doing this 5k times would take approx 25s

test/tangle-prune.test.js
```diff
@@ -0,0 +1,87 @@
...
+              return res(msg)
+            }
+          )
+        })
+    )
+
+    //console.time('publish')
+    Promise.all(publishes)
+      .then(async (msgs) => {
+        //console.timeEnd('publish')
```
Powersource, line 81: approx 4s. Most messages seem to only have 1 entry in the `previous` array

test/tangle-prune.test.js
arj03: nice

lib/prune-publish.js
```diff
@@ -0,0 +1,22 @@
...
+// SPDX-License-Identifier: LGPL-3.0-only
+
+module.exports = function prunePublish(ssb, content, cb) {
+  const group = content.tangles.group
+
+  ssb.db.create({ content, encryptionFormat: 'box2' }, (err, msg) => {
+    if (err) {
+      if (group.previous.length > 1) {
+        const halfLength = Math.ceil(group.previous.length / 2)
+        group.previous = group.previous.slice(0, halfLength)
```
mixmix, line 12: random request - can you shuffle the previous before slicing?

This means that if we hit problems here different messages which converge the tangles differently and it _should_ converge faster. If you slice and always take the first 63 it might be slow...

test/tangle-prune.test.js
```diff
@@ -0,0 +1,77 @@
+// SPDX-FileCopyrightText: 2022 Mix Irving
```
mixmix, line 1: rename file `test/lib/prune-publish.test.js`

mixmix (changes requested):
:heart: 

Nice succinct solution + tests, great work.
One minor request, could be a pre-optimisation but keen to have in in as had similar behaviour previously and worked well

lib/prune-publish.js
```diff
@@ -0,0 +1,22 @@
+// SPDX-FileCopyrightText: 2022 Andre 'Staltz' Medeiros <contact@staltz.com>
+//
+// SPDX-License-Identifier: LGPL-3.0-only
+
+module.exports = function prunePublish(ssb, content, cb) {
+  const group = content.tangles.group
+
+  ssb.db.create({ content, encryptionFormat: 'box2' }, (err, msg) => {
+    if (err) {
```
mixmix, line 9: we could actually check the error message includes "Too large" or something, but what you've done is fine

lib/prune-publish.js
mixmix:
```suggestion
        group.previous = group.previous
          .sort((a, b) => Math.random() < 0.5 ? -1 : 1)
          .slice(0, halfLength)
```

Powersource: It gets shuffled in this PR in lib/add-group-tangle.js

lib/prune-publish.js
Powersource: yeah but it'd have to work for all feed formats and then i'd have to find a spec somewhere and update it and eeh

test/tangle-prune.test.js
Powersource: done

lib/prune-publish.js
staltz: Create an issue for it, that would make us more comfortable merging this :)

Powersource: made one https://github.com/ssbc/ssb-tribes2/issues/25

mixmix (approved):
# :t-rex: 

NICE

staltz:
Can we get tests passing before merging?

PR #20, opened 2022-10-24
Fix failing tests caused by replication being fast

Powersource:
Prompted by https://github.com/ssbc/ssb-tribes2/pull/16

### DIFF
--- a/index.js
+++ b/index.js
@@ -6,7 +6,7 @@ const { promisify } = require('util')
 const pull = require('pull-stream')
 const paraMap = require('pull-paramap')
 const pullAsync = require('pull-async')
-const lodashGet = require('lodash.get')
+const lGet = require('lodash.get')
 const {
   where,
   and,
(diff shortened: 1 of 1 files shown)


staltz:
I can see what's going on: if we use db2 queries as the source of truth, then we know we're always up-to-date whenever querying (in  `list()` or `get()` or whatever), BUT then decryption doesn't actually work because [ssb-box2 looks up only the keys that were saved in ssb-keyring](https://github.com/ssbc/ssb-box2/blob/7f27034abc25e2d0b211b022bef08d6d2b46917e/format.js#L166). This means we need to reintroduce `ssb.box2.add*` APIs. If decryption doesn't work, then when you fetch messages, you get a string instead of the object, which is why the tests are failing.

It's a bit of a chicken and egg situation. But I think it could be better to bring back `start()` and just find a way of waiting for `start()` to process everything before continuing the logic inside the queries (`list()`, `get()`, etc).

I think we should pair code this one. Might be the quickest way of transferring thoughts.

mixmix:
@staltz the usual flow will be : 
1. you replicate a `group/add-member` message and decrypt it, discovering you have access to a new group
2. you add that group to your keyring, (create a new group subfeed,) register new subfeeds for replication,  and trigger a rebuild (ideally of just one group's metafeed leaves) 
3. you run the query (ssb.db.query, ssb.db.get?), which **waits** because reindexing is happening
4. you get the decrypted message

Powersource:
fixed with https://github.com/ssbc/ssb-tribes2/pull/26 instead

PR #24, opened 2022-10-25
add internal onCaughtUp API

staltz:
## Context

This is an alternative to #20.

## Problem

There is a race condition between the processing of messages in `start()` and the `list()` API. `list()` is _supposed_ to have all the up-to-date information, but by just querying ssb-box2 (and ssb-keyring), it doesn't, because there may be a message from the live query in `start()` that still hasn't trickled down and made its way to ssb-box2 (and ssb-keyring).

## Solution

We figured (pair programming with @Powersource) that we don't need to strictly solve the race condition in `list()` if we embrace eventual consistency in the apps, but in tests, we need to make strictly solve them, to make the correct assertions.

This PR adds `onCaughtUp(cb)` as an internal API just for our tests. The way it works is a bit hacky, I wanted to use `ssb.db.prepare()` (which in turn is just a light wrapper for jitdb's `prepare()`) but the problem is that `prepare()` only syncs the bitvector indexes, it doesn't fetch the messages from the log (which is an asynchronous operation with the file system). That meant that the processing in `start()` is "slower" than jitdb `prepare()` which would meant that we are not yet caught up.

That's why I added the comment:

```
// This is a bit hacky, because it fixes the race condition with `start()`
// by relying on ssb-db2 internals. This only works because the query in
// `start()` was initiated before this one, and because ssb-db2 and jitdb
// have an internal queue of queries that are executed in order.
```

@arj03 Since you're good at race conditions, and since this is a hacky solution, I'd really appreciate your thoughts on this PR.

### DIFF
--- a/index.js
+++ b/index.js
@@ -13,6 +13,7 @@ const {
   isDecrypted,
   type,
   live,
+  descending,
   toPullStream,
 } = require('ssb-db2/operators')
 const { keySchemes } = require('private-group-spec')
@@ -185,6 +186,7 @@ module.exports = {
       )
     }
 
+    let processingGroupAddMember = 0
     // Listeners for joining groups
     function start() {
       pull(
@@ -197,6 +199,7 @@ module.exports = {
           // TODO: call ssb-db2 reindexEncrypted
 
           if (lodashGet(msg, 'value.content.recps', []).includes(ssb.id)) {
+            processingGroupAddMember++
             const groupRoot = lodashGet(msg, 'value.content.root')
             const groupKey = lodashGet(msg, 'value.content.groupKey')
             const groupId = lodashGet(msg, 'value.content.recps[0]')
@@ -216,6 +219,7 @@ module.exports = {
                   root: groupRoot,
                 })
               }
+              processingGroupAddMember--
               return cb()
             })
           } else {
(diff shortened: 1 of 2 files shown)


arj03:
Hmm. This doesn't completely guarantee that you get all groups. Why not do something like check on `ssb.box2.listGroupIds` in a loop and if size > 0 and the last 2 ticks gave the same result, then we can assume that we got all. Or even better allow one to specify the number of groups that must exist.

staltz:
> Why not do something like check on `ssb.box2.listGroupIds` in a loop and if size > 0 and the last 2 ticks gave the same result

Hmm, that wouldn't guarantee anything, it all depends on how quick these ticks are, which isn't much better than just putting a `setTimeout(fn, 2000)`.

> Or even better allow one to specify the number of groups that must exist.

This could work, and it would be deterministic, but it affects the public API of the methods. Perhaps we should use `opts._expect` or some other name that starts with `_` to indicate it's an internal argument.

What do you think @Powersource ? Do you know how to implement that? If so, you can make another PR as an alternative to this one, and we can then compare.

Powersource:
> This doesn't completely guarantee that you get all groups.

Why?

> Or even better allow one to specify the number of groups that must exist.

That feels like it cheats around the test

Powersource:
@arj03

Powersource:
@staltz @arj03 

> Because you are testing it to be 0. It could be 0 after processed the first of 2 messages. Or even just in the beginning.

ah yeah it feels like there are a bunch of race conditions around that incrementor

> Perhaps we should use opts._expect

Hmm so if we do that then e.g. `list()` when we provide `_expect` would have to retry `ssb.box2.listGroupIds` in some kind of loop. But then the logic of `list()` is pretty different when run in tests compared to when run in production? Or is there a less invasive way of making it wait for the correct number of list items?

staltz:
What if we do the retry logic in the tests? Using async-await, you can use a for loop (or while loop) where each iteration calls `list()` and checks the amount of results, and retries if it's still not done. The for-loop can have a limited number of tries, but a large number of tries (say 20) just to ensure that even with absurd waiting times the tests still pass.

arj03:
Yeah maybe better, otherwise we might need a listGroupIds that is live.

Powersource:
boom mic drop https://github.com/ssbc/ssb-tribes2/pull/26

Powersource:
fixed with https://github.com/ssbc/ssb-tribes2/pull/26 instead

Issue #13, opened 2022-10-13
Cloaked links should be URIs not sigils


Powersource:
It's the future :sparkles: 

Made some related issues
https://github.com/ssbc/private-group-spec/issues/17
https://github.com/ssbc/ssb-uri-spec/issues/15

Powersource:
Could be nice and clean if we could uninstall ssb-ref from this module

Powersource:
Latest discussion https://github.com/ssbc/ssb-uri-spec/issues/18

Powersource:
And pr in the spec https://github.com/ssbc/private-group-spec/pull/18

PR #29, opened 2022-11-02
Use URIs instead of sigils

Powersource:
Fixes https://github.com/ssbc/ssb-tribes2/issues/13

* [x] Maybe do https://github.com/ssbc/ssb-uri2/issues/18 before merging to make the code here cleaner
  * [x] If we do the above then uninstall ssb-ref again
* [x] Should we update `buildGroupId` to not calculate on sigil keys? I guess? no. see comment below
  * [x] We might need to update the envelope spec? Depends on how `encode` works i guess? What's SLP? https://github.com/ssbc/envelope-spec/tree/master/cloaked_msg_id

### DIFF
--- a/index.js
+++ b/index.js
@@ -22,7 +22,8 @@ const {
   },
 } = require('private-group-spec')
 const { SecretKey } = require('ssb-private-group-keys')
-const { fromMessageSigil } = require('ssb-uri2')
+const { fromMessageSigil, fromFeedSigil } = require('ssb-uri2')
+const ref = require('ssb-ref')
 //const Crut = require('ssb-crut')
 const buildGroupId = require('./lib/build-group-id')
 const AddGroupTangle = require('./lib/add-group-tangle')
@@ -125,7 +126,11 @@ module.exports = {
       get(groupId, (err, { secret, root }) => {
         if (err) return cb(err)
 
-        const recps = [groupId, ...feedIds]
+        const feedIdUris = feedIds.map((feedId) =>
+          ref.isFeedId(feedId) ? fromFeedSigil(feedId) : feedId
+        )
+
+        const recps = [groupId, ...feedIdUris]
 
         const content = {
           type: 'group/add-member',
@@ -205,7 +210,11 @@ module.exports = {
         pull.asyncMap((msg, cb) => {
           // TODO: call ssb-db2 reindexEncrypted
 
-          if (lodashGet(msg, 'value.content.recps', []).includes(ssb.id)) {
+          if (
+            lodashGet(msg, 'value.content.recps', []).includes(
+              fromFeedSigil(ssb.id)
+            )
+          ) {
             const groupRoot = lodashGet(msg, 'value.content.root')
             const groupKey = lodashGet(msg, 'value.content.groupKey')
             const groupId = lodashGet(msg, 'value.content.recps[0]')
--- a/index.js
+++ b/index.js
@@ -22,6 +22,7 @@ const {
   },
 } = require('private-group-spec')
 const { SecretKey } = require('ssb-private-group-keys')
+const { fromMessageSigil } = require('ssb-uri2')
 //const Crut = require('ssb-crut')
 const buildGroupId = require('./lib/build-group-id')
 const AddGroupTangle = require('./lib/add-group-tangle')
@@ -70,7 +71,7 @@ module.exports = {
           const data = {
             id: buildGroupId({ groupInitMsg, groupKey: groupKey.toBuffer() }),
             secret: groupKey.toBuffer(),
-            root: groupInitMsg.key,
+            root: fromMessageSigil(groupInitMsg.key),
             subfeed: '',
           }
 
--- a/index.js
+++ b/index.js
@@ -22,9 +22,7 @@ const {
   },
 } = require('private-group-spec')
 const { SecretKey } = require('ssb-private-group-keys')
-const { fromMessageSigil, fromFeedSigil } = require('ssb-uri2')
-const ref = require('ssb-ref')
-//const Crut = require('ssb-crut')
+const { fromMessageSigil, fromFeedSigil, isFeedSSBURI } = require('ssb-uri2')
 const buildGroupId = require('./lib/build-group-id')
 const AddGroupTangle = require('./lib/add-group-tangle')
 const prunePublish = require('./lib/prune-publish')
@@ -126,9 +124,11 @@ module.exports = {
       get(groupId, (err, { secret, root }) => {
         if (err) return cb(err)
 
-        const feedIdUris = feedIds.map((feedId) =>
-          ref.isFeedId(feedId) ? fromFeedSigil(feedId) : feedId
-        )
+        const feedIdUris = feedIds
+          .map((feedId) =>
+            isFeedSSBURI(feedId) ? feedId : fromFeedSigil(feedId)
+          )
+          .filter(Boolean)
 
         const recps = [groupId, ...feedIdUris]
 
--- a/lib/add-group-tangle.js
+++ b/lib/add-group-tangle.js
@@ -2,8 +2,8 @@
 //
 // SPDX-License-Identifier: LGPL-3.0-only
 
-const { isCloakedMsg } = require('ssb-ref')
 const set = require('lodash.set')
+const { isIdentityGroupSSBURI } = require('ssb-uri2')
 const GetGroupTangle = require('./get-group-tangle')
 
 module.exports = function AddGroupTangle(server) {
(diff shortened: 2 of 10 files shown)


lib/build-group-id.js
```diff
@@ -33,9 +33,12 @@ module.exports = function buildGroupId({
     }
   }
 
-  const cloakedMsgId = new CloakedMsgId(msgId, readKey).toString()
+  const cloakedMsgId = new CloakedMsgId(msgId, readKey)
+    .toString()
+    .replaceAll('/', '_')
+    .replaceAll('+', '-')
```
staltz, line 39: Can we use https://github.com/joaquimserafim/base64-url `base64url.escape()`? I'm afraid that the more we sprinkle these replaceAll (across various repos), the more likely is that one day one of them will get a tiny detail wrong.

lib/add-group-tangle.js
```diff
@@ -15,15 +15,10 @@ module.exports = function AddGroupTangle(server) {
...
+    if (!isIdentityGroupSSBURI(content.recps[0])) return cb(null, content)
 
     getGroupTangle(content.recps[0], (err, tangle) => {
-      // NOTE there are two ways an err can occur in getGroupTangle
-      // 1. recps is not a groupId
-      // 2. unknown groupId,
-
-      // Rather than cb(err) here we we pass it on to boxers to see if an err is needed
-      if (err) return cb(null, content)
+      if (err) return cb(err)
```
Powersource, line 21: Took the liberty to actually error here. When validating against the spec we don't allow for tangle-less messages anyway so I think this comment probably isn't relevant here anymore.

staltz: Good call, also the comment spoke about "boxers" and we don't use that anymore, it's a db1 thing

package.json
```diff
@@ -49,7 +51,7 @@
     "pretty-quick": "^3.1.3",
     "secret-stack": "^6.4.1",
     "ssb-caps": "^1.1.0",
-    "ssb-db2": "^6.2.3",
+    "ssb-db2": "ssbc/ssb-db2#uri-get",
```
Powersource, line 54: Needs merge too https://github.com/ssbc/ssb-db2/pull/401

Powersource:
The `buildGroupId` situation seems to be solved since it looks agnostic to whether it's a sigil or a uri

```js
  const msgId = bfe.encode(groupInitMsg.key) // this line is from `buildGroupId`
  const msgIdUri = bfe.encode(uri.fromMessageSigil(groupInitMsg.key))

  console.log('are they equal', msgId.equals(msgIdUri)) // prints true
```

staltz:
@Powersource Could we keep classic refs as sigils? I totally back the idea of moving from sigil cloaked to URI cloaked, but classic feed IDs and msg IDs IMO should stay as sigils, for consistency with everything else (e.g. `sbot.id` is a sigil).

Powersource:
> Could we keep classic as sigils?

Why were classic uris introduced in that case? Shouldn't sigils be included in uri-spec if we want to keep using them? Also this discussion is better to have over here https://github.com/ssbc/private-group-spec/pull/18 since here I'm just adapting to that spec.

Powersource:
Also note that the api here still should accept both sigils and uris. The adaptation is mostly internal datastructures and stuff. We use helper functions in the tests but that's just because we want to do deepEquals and stuff.

lib/build-group-id.js
Powersource: switched to the library now. but had to add a = at the end because the library removed the padding

Powersource: actually there's this in uri2 but it's not exported https://github.com/ssbc/ssb-uri2/blob/ab051cbe99a7b87027305464d8ac6cb65c24943b/src/index.ts#L33-L41

package.json
```diff
@@ -32,15 +32,17 @@
...
-    "ssb-bfe": "^3.6.0",
-    "ssb-box2": "^3.0.0",
+    "ssb-bfe": "^3.7.0",
+    "ssb-box2": "^3.0.1",
     "ssb-crut": "^4.6.1",
     "ssb-meta-feeds": "~0.30.0",
-    "ssb-private-group-keys": "^1.0.0"
+    "ssb-private-group-keys": "^1.0.0",
+    "ssb-ref": "^2.16.0",
+    "ssb-uri2": "^2.4.0"
```
Powersource, line 45: Need to update after https://github.com/ssbc/ssb-uri2/pull/19

Would use the branch already but i don't know how to make ts play nice

staltz:
> Why were classic uris introduced in that case?

As a shortcut to sigils. Other than that, classic URIs are not supposed to be in the data layer. E.g. `msg.value.author` in a classic message has to be a sigil `@`. If it's an SSB URI `ssb:feed/classic/__` then signatures are going to look different, and everything else is going to break.

Powersource:
replying in the spec pr

index.js
```diff
@@ -124,7 +124,13 @@ module.exports = {
       get(groupId, (err, { secret, root }) => {
         if (err) return cb(err)
 
-        const recps = [groupId, ...feedIds]
+        const feedIdUris = feedIds
+          .map((feedId) =>
+            isFeedSSBURI(feedId) ? feedId : fromFeedSigil(feedId)
```
mixmix, line 129: what happens if I pass in the feedId `'dog'` ... it's not an ssb-uri, so we try to convert it. But it fails... how does it fail? is it null, does it throw.

Either way, I want that to be caught and for encryption to fail? I think that might be covered by like ssb-box2 though. 
ok.... lets leave it to the encryption to catch invalid inputs... hmm yes

index.js
```diff
@@ -204,7 +210,11 @@ module.exports = {
         pull.asyncMap((msg, cb) => {
           // TODO: call ssb-db2 reindexEncrypted
 
-          if (lodashGet(msg, 'value.content.recps', []).includes(ssb.id)) {
+          if (
+            lodashGet(msg, 'value.content.recps', []).includes(
+              fromFeedSigil(ssb.id)
```
mixmix, line 215: probably don't need to calculate this every time, could extract variable.
Also this is gonna change with meta-feeds - this needs to be the root id. (ssb.id is the `main` leaf id)

lib/add-group-tangle.js
mixmix: uhhh.. I think we use boxers still, we just register them in plugins like ssb-box and ssb-box2 right?
Anyway, we just need to be clear about who catches encryption errors

@staltz (or anyone else) - do you agree that it's if there are 7 recps and one of them is bung, the encryption should fail (NOT the alternative which would be that it succeeds but only encrypts to 6, and you just don't know it wasn't encrypted to the 7th)

lib/build-group-id.js
```diff
@@ -33,9 +34,11 @@ module.exports = function buildGroupId({
     }
   }
 
-  const cloakedMsgId = new CloakedMsgId(msgId, readKey).toString()
+  const cloakedMsgId = base64url.escape(
+    new CloakedMsgId(msgId, readKey).toString()
+  )
```
mixmix, line 39: cc @staltz can you speak this this? these are different functions but I don't know if it matters.
This module `base64url.escape` does: https://github.com/joaquimserafim/base64-url/blob/master/index.js#L10-L20
```js
function unescape (str) {
  return (str + '==='.slice((str.length + 3) % 4))
    .replace(/-/g, '+')
    .replace(/_/g, '/')
}

function escape (str) {
  return str.replace(/\+/g, '-')
    .replace(/\//g, '_')
    .replace(/=/g, '')
}
```
# :hot_pepper:  DOES THE `=` MATTER?

Meanwhile `ssb-uri2` does: https://github.com/ssbc/ssb-uri2/blob/main/src/index.ts#L33-L41
```js
const Base64 = {
  unsafeToSafe(input: string) {
    return input.replace(/\+/g, '-').replace(/\//g, '_');
  },

  safeToUnsafe(input: string) {
    return input.replace(/-/g, '+').replace(/_/g, '/');
  },
};
```

mixmix: I think these functions are functionally the same. the = is filler right

mixmix: yeah if you're already using ssb-uri2 you should expose that escape unescape function and use that

test/add-member.test.js
```diff
@@ -54,7 +55,10 @@ test('add member', async (t) => {
     const group = await kaitiaki.tribes2.create()
     t.true(group.id, 'creates group')
 
-    const authorIds = [newPerson.id, ssbKeys.generate().id]
+    const authorIds = [
+      fromFeedSigil(newPerson.id),
+      fromFeedSigil(ssbKeys.generate().id),
+    ]
```
mixmix, line 61: :hot_pepper: I think you want to test it can take one of each and not panic right?

Your code I read I think maps them into URIs anyway right?

mixmix (approved):
This is a sweet change. I'm so pumped for all ssb-uris! such consistency

sorry was thinking out loud on some of these comments.
FYI I use the emoji
- :fire: = this is for sure gonna catch fire, this must be resolved
- :hot_pepper: = I think could be a problem, might be wrong

All other comments are just estabilishing shared norms, sharing learning, asking about something I'm keen to learn about, or just yelling "HECK YEAH". (you can ignore these if you want)

index.js
Powersource: haven't checked now but i think it outputs null which in turn will get caught by schema checking

index.js
Powersource: i think andre is changing this in the metafeeds pr

Powersource:
a lot of comments but approved so i'll take it as fine to merge?

Issue #27, opened 2022-10-28
Match all publishes against spec


Powersource:
We could easily adapt `private-group-spec/group/add-member/schema.json` to check the tangle on a normal message. Then we could check everything being published by `publish`, including the work that the tangle pruning does (that sort of gets skipped by the addMember spec being checked before publish/tanglePrune

Powersource:
Just spent waay too long trying to get external schemas working with is-my-ssb/json-valid. My conclusion is that it basically doesn't work or at least doesn't have any bloody documentation uurgh. I'll just hardcode the duplicated parts, alternatively manually splice json together from different files.

Powersource:
pr in the spec https://github.com/ssbc/private-group-spec/pull/21

needs to be followed up with a pr in tribes2 that actually uses that schema

Powersource:
spec pr merged

Issue #17, opened 2022-10-18
Test reindexEncrypted messages when you get added to a new group


Powersource:
https://github.com/ssbc/ssb-tribes2/blob/c4dcfd1c4f9a78009f1cb6b6b61340349cb59d10/index.js#L197

For when you get added late to a group, and you already technically have messages from the group

Powersource:
we call reindexencrypted but we should test it

Powersource:
We add a new test for this in https://github.com/ssbc/ssb-tribes2/pull/47/files#diff-427320432f1167b8bdddd47082855ab955cd0fa2ffce8acdac14e2fda3b975be

PR #34, opened 2022-11-08
use ssb-meta-feeds according to group spec

staltz:
Spiritual successor to the [meta-feeds](https://github.com/ssbc/ssb-tribes2/compare/meta-feeds) branch.

Implements the [metafeeds group spec](https://github.com/ssbc/ssb-meta-feed-group-spec).

Asking for a review from @Powersource.

And for visibility, pinging @mixmix .


index.js
```diff
@@ -54,36 +70,48 @@ module.exports = {
...
+          },
+          (err, groupInitMsg) => {
             if (err) return cb(err)
 
-            return cb(null, data)
-          })
-        }
-      )
+            const data = {
+              id: buildGroupId({ groupInitMsg, groupKey: groupKey.toBuffer() }),
```
mixmix, line 93: does this now return an ssb-uri

index.js
```diff
@@ -54,36 +70,48 @@ module.exports = {
...
+              subfeed: groupFeed.keys,
+            }
+
+            ssb.box2.addGroupInfo(data.id, {
+              key: data.secret,
+              root: data.root,
+            })
+
+            // Adding myself for recovery reasons
+            addMembers(data.id, [ssb.id], {}, (err) => {
```
mixmix, line 105: :fire: you are adding the wrong "person" 
addMembers needs to take rootFeedIds, and this is then translated encryptionKey for the `invitations` leaf (this is the spec I need to finish writing), and published in the in the `invitations` feed.

I think we need to specify in the README which keys should go into addMembers!

index.js
```diff
@@ -151,14 +179,20 @@ module.exports = {
...
 
-        publish(content, (err, msg) => {
+        const invitations = {
+          purpose: 'invitations',
+          feedFormat: 'classic',
+        }
+
+        ssb.metafeeds.findOrCreate(invitations, (err, invitationsFeed) => {
           if (err) return cb(err)
+          ssb.box2.addKeypair(invitationsFeed.keys)
```
mixmix, line 189: :fire: I think it's 
```suggestion
          ssb.box2.addKeypair(invitationsFeed.metadata.encryptionKeys)
```

index.js
```diff
@@ -171,11 +205,20 @@ module.exports = {
       const recps = content.recps
       if (!recps || !Array.isArray(recps) || recps.length < 1)
         return cb(new Error('Missing recps'))
+      const groupId = recps[0]
```
mixmix, line 208: we should check more - for this the recps length should be EXACTLY 1, and the should be a URI that `startsWith('ssb:identity/group/')` or whatever we settled on

index.js
```diff
@@ -31,21 +30,38 @@ module.exports = {
...
+      const groupFeedDetails = {
+        purpose: groupKey.toString(),
+        feedFormat: 'classic',
+        recps,
+        encryptionFormat: 'box2',
+      }
+
+      ssb.metafeeds.findOrCreate(groupFeedDetails, (err, groupFeed) => {
+        if (err) return cb(err)
+        ssb.box2.addKeypair(groupFeed.keys)
```
mixmix, line 57: TODO: change

index.js
```diff
@@ -54,36 +70,48 @@ module.exports = {
...
-      )
+            const data = {
+              id: buildGroupId({ groupInitMsg, groupKey: groupKey.toBuffer() }),
+              secret: groupKey.toBuffer(),
+              root: groupInitMsg.key,
+              subfeed: groupFeed.keys,
+            }
+
+            ssb.box2.addGroupInfo(data.id, {
+              key: data.secret,
```
mixmix, line 100: `<aside>` if we're breaking the API of box2, I would love to rename the groupKey to `secret` or `groupSecret` everywhere... or as many places as possible. I really regret the over-use of the word "Key"

mixmix (changes requested):
sorry this is a spicy poo review

index.js
staltz: other PR

index.js
Powersource: there's this issue for that https://github.com/ssbc/ssb-tribes2/issues/27 which is probably good to do soon/maybe even before this pr so we know we don't break stuff

index.js
```diff
@@ -31,21 +30,38 @@ module.exports = {
...
+      const groupFeedDetails = {
+        purpose: groupKey.toString(),
+        feedFormat: 'classic',
+        recps,
+        encryptionFormat: 'box2',
+      }
+
+      ssb.metafeeds.findOrCreate(groupFeedDetails, (err, groupFeed) => {
+        if (err) return cb(err)
+        // ssb.box2.addKeypair(groupFeed.keys)
```
Powersource, line 57: ?

staltz: I thought I had deleted it, but apparently I commented it. Will delete.

index.js
Powersource: yeah so `recps` in the addMembers message should be the root ids but the actual encryption shouldn't look at those keys like it usually does :dizzy_face:

index.js
```diff
@@ -201,34 +243,29 @@ module.exports = {
           live({ old: true }),
           toPullStream()
         ),
+        pull.filter((msg) =>
+          lodashGet(msg, 'value.content.recps', []).includes(ssb.id)
+        ),
         pull.asyncMap((msg, cb) => {
-          // TODO: call ssb-db2 reindexEncrypted
```
Powersource, line 205: we should add tests for this

staltz: Maybe I should just remove reindexEncrypted from this PR then.

Powersource: yeah we have an issue for this, maybe better to do it all later https://github.com/ssbc/ssb-tribes2/issues/17

index.js
staltz: The emoji :dizzy_face: describes me well these days

mixmix:
```suggestion
            addMembers(data.id, [rootMF.keys.id], {}, (err) => {
```

index.js
```diff
@@ -159,14 +187,20 @@ module.exports = {
         if (!addMemberSpec(content))
           return cb(new Error(addMemberSpec.errorsString))
 
-        publish(content, (err, msg) => {
+        const invitations = {
+          purpose: 'invitations',
+          feedFormat: 'classic',
```
mixmix, line 192: :fire:
this feed needs DH keys attached in metadata

mixmix, line 192:
```suggestion
          feedFormat: 'classic',
          metadata: {
            encryption: { key: <base64>, format: 'curve25519' }
          }
```

staltz: We need a spec for this. Or, we need this put in some existing spec.

staltz: Also, how do I convert from this to a `DHKeys`? It seems like DHKeys needs `{ public, secret }`

Powersource:
strange, in this pr the tests stopped working in node 19 for me even though they've worked before. they just freeze when starting to run. node 16 works for me though

index.js
mixmix: We are not converting, these will be raw DH keys. And we are only publishing the public part!!!  the private part must be persisted by you

staltz: https://github.com/ssbc/ssb-meta-dm-spec/issues/1

staltz:
Ready for review, and I think this should be the last code PR to merge before milestone completion, because https://github.com/ssbc/ssb-meta-dm-spec/issues/1

@arj03

index.js
```diff
@@ -54,36 +84,55 @@ module.exports = {
...
+                if (err) return cb(err)
+
+                return cb(null, data)
+              })
+            }
+          )
+        })
+      })
+
+      // TODO: consider what happens if the app crashes between any step
```
arj03, line 135: :) yep was wondering that

index.js
```diff
@@ -144,29 +185,30 @@ module.exports = {
...
 
-          //TODO: this is an optimization, use the db query instead for now
-          //keystore.group.registerAuthors(groupId, feedIds, (err) => {
-          //  if (err) return cb(err)
-          return cb(null, msg)
-          //})
+          addGroupTangle(content, (err, content) => {
+            if (err) return cb(err)
+
+            prunePublish(ssb, content, invitationsFeed.keys, cb)
```
arj03, line 210: can we call this publishAndPrune?

arj03 (approved):
Just some minor things I noticed, otherwise looks good and good to have this into a shape where it mostly works :)

index.js
arj03: wait, I don't understand that function. It seems to loop, so it will do many creates? I thought it would prune down the tangle and then publish.

arj03: Then forget all of this :)

index.js
staltz: Made a task for this

index.js
```diff
@@ -203,47 +254,46 @@ module.exports = {
...
+                return
               }
 
               if (!info) {
-                // we're not already in the group
+                // We're not yet in the group
                 ssb.box2.addGroupInfo(groupId, {
-                  key: groupKey,
-                  root: groupRoot,
+                  key: lodashGet(msg, 'value.content.groupKey'),
```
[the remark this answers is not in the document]
staltz: Made a task for this

index.js
staltz: Done

staltz:
Thanks for the review!

PR #35, opened 2022-11-09
fix the max add-member

mixmix:
The max members you can add at a time is 15:
group/add-member message recps: [groupId, ...feedIds] and max recps is 16


staltz (approved):
Yup

staltz:
Good morning!

Issue #40, opened 2022-12-07
Crash resistant create() API


staltz:
We should "restart" any create() that was partially done but interrupted by a crash.

Plus tests.

-------

* [x] https://github.com/ssbc/ssb-tribes2/pull/56
* [x] Check/update the specs
* [x] https://github.com/ssbc/ssb-meta-feeds-group-spec/pull/15
* [x] https://github.com/ssbc/private-group-spec/pull/24
* [x] Update spec to include a copy of the group key in the group init. https://github.com/ssbc/private-group-spec/pull/25
* [x] Update code with group key in init. https://github.com/ssbc/ssb-tribes2/pull/62

Powersource:
@staltz @mixmix @arj03 is there any prior art on this?

Powersource:
if this crashes on line 89, then we've got a new group feed that won't be used. that doesn't break anything since we'll just create a new one the next time we try to create a group, but it's slightly inefficient. does that need to be fixed?

https://github.com/ssbc/ssb-tribes2/blob/206138ecabb88f0f807e4d3e2218808d53db79ab/index.js#L83-L95

Powersource:
i guess it's a similar problem to, and maybe has the same solution as, when you crash inbetween creating the root msg and inviting yourself to the group. which similarly should mostly just be suboptimal

Powersource:
i was gonna ask if this was redundant

https://github.com/ssbc/ssb-tribes2/blob/206138ecabb88f0f807e4d3e2218808d53db79ab/index.js#L56

since we later add ourselves to the group anyway. but maybe encrypting to ourselves here as well could be a way of saving the group early

or wait, when recovering from scratch, will you need the feed announcement encrypted to yourself to be able to find the feed again? hmm maybe not, since we do invite ourselves on our invitations feed, and we know how to find the invitations feed

staltz:
> if this crashes on line 89, then we've got a new group feed that won't be used. that doesn't break anything since we'll just create a new one the next time we try to create a group

I guess it's fine, but it depends on what kind of crash it is. If it's a totally random crash, then the issue of leaving empty groups behind is definitely not a problem because it's very rare. But let's say that this is a deterministic crash, e.g. every time the user creates a group, it always crashes on line 89. Then crash-resistance is important, otherwise we're creating a lot of group feeds in vain.

[...]

PR #44, opened 2022-12-19
Recover if crashed after creating a group feed

Powersource:
for https://github.com/ssbc/ssb-tribes2/issues/40

The "crash on line 89" discussed in there


test/create.test.js
```diff
@@ -112,3 +111,85 @@ test.skip('tribes.create (opts.addPOBox)', (t) => {
     t.end()
   })
 })
+
+test('create reuses an unused group feed (because of an earlier crash or something)', (t) => {
+  const server = Testbot()
+
+  function countGroupFeeds(cb) {
+    pull(
+      server.metafeeds.branchStream({ live: false }),
```
Powersource, line 120: branchStream currently isn't outputting anything, even when i know i have at least a root and group feed

staltz: The default value for `old` is false https://github.com/ssbc/ssb-meta-feeds/blob/58454e8440eab0bd64ed0bf63c6cd9a9c8aaf5d1/lookup.js#L306-L309 so in this case you have both live and old false, which means it'll return pull.empty() https://github.com/ssbc/ssb-meta-feeds/blob/58454e8440eab0bd64ed0bf63c6cd9a9c8aaf5d1/lookup.js#L336

index.js
```diff
@@ -48,12 +49,42 @@ module.exports = {
       ssb.metafeeds.findOrCreate(details, cb)
     }
 
-    function findOrCreateGroupFeed(input, cb) {
-      const secret = Buffer.isBuffer(input) ? new SecretKey(input) : input
+    function findEmptyGroupFeed(rootId, cb) {
+      console.log('in empty function')
+      pull(
+        ssb.metafeeds.branchStream({ root: rootId, live: false }),
```
staltz, line 55: Add `old` opt here

index.js
```diff
@@ -48,12 +49,42 @@ module.exports = {
...
+            })
+          )
+        }),
+        pull.filter(),
+        pull.unique('id'),
+        //pull.flatten(),
+        pull.collect((err, empties) => {
+          if (err) return cb(err)
+          return cb(null, empties[0])
+        })
```
staltz, line 79: This pull.collect could be replaced with `pull.take(1)` and `pull.drain`, because pull.collect can be bad for memory consumption.

index.js
```diff
@@ -48,12 +49,42 @@ module.exports = {
...
 
-    function findOrCreateGroupFeed(input, cb) {
-      const secret = Buffer.isBuffer(input) ? new SecretKey(input) : input
+    function findEmptyGroupFeed(rootId, cb) {
+      console.log('in empty function')
+      pull(
+        ssb.metafeeds.branchStream({ root: rootId, live: false }),
+        pull.filter((branch) => branch.length === 4),
+        pull.map((branch) => branch[3]),
+        pull.filter((feed) => feed.recps),
```
staltz, line 58: This filter could look into `feed.purpose` and test whether it matches something that looks like a group secret.

index.js
```diff
@@ -48,12 +49,42 @@ module.exports = {
...
+          return cb(null, empties[0])
+        })
+      )
+    }
 
+    function findOrCreateFromSecret(secret, rootId, cb) {
       const recps = [
         { key: secret.toBuffer(), scheme: keySchemes.private_group },
-        // TODO: add self to recps, for crash-resistance
+        // encrypt to myself to be able to get back to the group without finding the group/add-member for me (maybe i crashed before adding myself)
```
staltz, line 86: Is this comment a TODO?

index.js
```diff
@@ -80,14 +141,16 @@ module.exports = {
...
-      const secret = new SecretKey()
-
       ssb.metafeeds.findOrCreate(function gotRoot(err, root) {
         if (err) return cb(err)
 
-        findOrCreateGroupFeed(secret, function gotGroupFeed(err, groupFeed) {
+        findOrCreateGroupFeed(null, function gotGroupFeed(err, groupFeed) {
           if (err) return cb(err)
 
+          console.log('create group feed', groupFeed)
```
staltz, line 150: The `groupFeed` may be empty in case there has been no crash. In that case we should call `findOrCreateGroupFeed(secret, ...`

test/create.test.js
staltz: One way to simulate a crash with a partial state is to manually create an empty group feed in the test code, instead of using ssb.tribes2.create.

index.js
Powersource: i don't think so

test/create.test.js
Powersource: > The default value for old is false

wow lmao it literally returns empty every time

> manually create an empty group feed in the test code

yep that's what i ended up doing

Powersource:
All tests are passing locally

index.js
Powersource: findOrcreateGroupFeed does that internally, see comment 2.2 inside it

Powersource:
Not sure what's up with CI

test/create.test.js
staltz: > wow lmao it literally returns empty every time

Perhaps it could return a `pull.error()` stream instead.

staltz:
Have you used the same version of Node.js locally? `16.x`

Powersource:
Ran 19 before but all tests passing on v16.19.0 as well

Powersource:
I think the thing that finally made the tests fail locally for me was deleting pnpm-lock.yaml (not just node_modules)

Powersource:
strange, works again when i run the file directly

staltz:
Tip: replace ubuntu-latest in CI with ubuntu-20.04 or whatever the string is to get ubuntu 20.04

staltz:
Didn't the tests just now pass fine?

Powersource:
yes because i've pinned meta feeds to .1

package.json
```diff
@@ -45,14 +45,15 @@
     "ssb-bfe": "^3.7.0",
     "ssb-box2": "^4.0.0",
     "ssb-crut": "^4.6.1",
-    "ssb-db2": "^6.2.6",
-    "ssb-meta-feeds": "^0.38.1",
+    "ssb-db2": "6.3.0",
+    "ssb-meta-feeds": "0.38.1",
```
Powersource, line 49: Upgrading to 0.38.2 breaks tests

```
get added to a group
    ✔ tribes2 started for both alice and bob
    ✔ alice and bob replicate their trees
    in empty function
    emptyFeed undefined
    about to post root msg
    just published root msg
    ✔ alice created a group
    ✔ alice added bob to the group
    ✖ 5) Error: retryUntil timed out
      Expected error to be falsy
      

  add member
```

index.js

index.js
Powersource: checking that the length is 44 now

package.json
Powersource: Sometimes it fails in this place, sometimes in the next test or later. It is really flaky.

@arj03 do you know what might be up here? this is the pr for .2 https://github.com/ssbc/ssb-meta-feeds/pull/112/files?diff=split&w=1

I tried to dig and get some more info using this

```
diff --git a/package.json b/package.json
index 3e16c57..86b5c1c 100644
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
     "ssb-box2": "^4.0.0",
     "ssb-crut": "^4.6.1",
     "ssb-db2": "6.3.0",
-    "ssb-meta-feeds": "0.38.1",
+    "ssb-meta-feeds": "0.38.2",
     "ssb-private-group-keys": "^1.1.1",
     "ssb-ref": "^2.16.0",
     "ssb-uri2": "^2.4.1"
diff --git a/test/helpers/replicate.js b/test/helpers/replicate.js
index 2338965..60ff5aa 100644
--- a/test/helpers/replicate.js
+++ b/test/helpers/replicate.js
@@ -4,6 +4,16 @@
 
 const { promisify: p } = require('util')
 const pull = require('pull-stream')
+const paraMap = require('pull-paramap')
+const {
+  where,
+  and,
+  isDecrypted,
+  type,
+  live,
+  author,
+  toPullStream,
+} = require('ssb-db2/operators')
 const deepEqual = require('fast-deep-equal')
 
 /**
@@ -31,6 +41,24 @@ module.exports = async function replicate(person1, person2, opts = {}) {
   let drain1
   pull(
     person1.metafeeds.branchStream({ old: true, live: true }),
+    paraMap((branch, cb) => {
+      if (
+        branch.length === 4 &&
+        branch[3].purpose === 'group/additions' &&
+        branch[3].seed !== null
+      ) {
+        console.log('person1 p1 additions feed', branch[3])
+        return pull(
+          person1.db.query(where(author(branch[3].id)), toPullStream()),
+          pull.collect((err, res) => {
+            console.log('p1 seeing p1 additions', res)
+
+            return cb(null, branch)
+          })
+        )
+      }
+      return cb(null, branch)
+    }),
     pull.flatten(),
     pull.map((feedDetails) => feedDetails.id),
     pull.unique(),
@@ -43,6 +71,24 @@ module.exports = async function replicate(person1, person2, opts = {}) {
   let drain2
   pull(
     person2.metafeeds.branchStream({ old: true, live: true }),
+    paraMap((branch, cb) => {
+      if (
+        branch.length === 4 &&
+        branch[3].purpose === 'group/additions' &&
+        branch[3].seed === null
+      ) {
+        console.log('person2 p1 additions feed', branch[3])
+        return pull(
+          person2.db.query(where(author(branch[3].id)), toPullStream()),
+          pull.collect((err, res) => {
+            console.log('p2 seeing p1 additions', res)
+
+            return cb(null, branch)
+          })
+        )
+      }
+      return cb(null, branch)
+    }),
     pull.flatten(),
     pull.map((feedDetails) => feedDetails.id),
     pull.unique(),
```

I thought that the problem might be that p2 gets p1's additions feed but doesn't manage to decrypt it, but p2 doesn't seem to replicate p1's additions feed even if p2 discovers it. I.e. it prints stuff like

```
p2 seeing p1 additions []
    p1 seeing p1 additions [
      {
        key: '%kcZefmGgRvoiiQgvW6PuipCA71XQf8DANVnJMYo6kX4=.sha256',
        value: {
...
```

-------

hmm although i write that, if i switch back to .1, my small script above still prints [] for p2 and the tests pass, so my script is buggy in some way i guess. gotta go eat though.

Powersource: moved to https://github.com/ssbc/ssb-tribes2/issues/48

index.js
```diff
@@ -48,12 +49,50 @@ module.exports = {
...
+            pull.take(1),
+            pull.collect((err, res) => {
+              if (res.length === 0) {
+                // we're only interested in empty feeds
+                return cb(null, feed)
+              } else {
+                return cb()
+              }
+            })
+          )
```
staltz, line 72: There's a better way to count the size of a feed that doesn't depend on actually retrieving the messages from the log (thus relying only on jitdb indexes) and without pull.collect:

```js
ssb.db.query(where(author(feed.id)), count(), toCallback(cb)
```

The `cb` is called as `cb(err, num)` where `num` is the number of messages matching the query. Then in the filter you can do `pull.filter(num => num === 0)`.

Well, it's not going to be that simple because you then have `pull.unique('id')`, but at least this gives a hint how to get it done

index.js
```diff
@@ -48,12 +49,50 @@ module.exports = {
...
+        pull.take(1),
+        pull.drain(
+          (empty) => {
+            found = true
+            return cb(null, empty)
+          },
+          (err) => {
+            if (err) return cb(err)
+            if (found) return
+            return cb(null, undefined)
```
staltz, line 85: These 3 lines look a bit confusing. I recommend

```js
if (err) cb(err)
else if (!found) cb()
```

index.js
```diff
@@ -69,6 +108,34 @@ module.exports = {
...
+            if (emptyFeed) {
+              return cb(null, emptyFeed)
+            } else {
+              const newSecret = new SecretKey()
+              findOrCreateFromSecret(newSecret, root.id, cb)
+            }
+          })
+        }
+      })
+    }
```
staltz, line 137: Neat implementation :+1:

test/create.test.js
```diff
@@ -112,3 +111,83 @@ test.skip('tribes.create (opts.addPOBox)', (t) => {
...
+            },
+            (err) => {
+              if (err) t.fail(err)
+
+              countGroupFeeds((err, num) => {
+                if (err) t.fail(err)
+                t.equal(
+                  num,
+                  2,
+                  'there is 1 used group feed and 1 dummy one, the empty one we created now'
```
staltz, line 168: It seems to me like there are actually 2 empty group feeds? The first group feed wasn't published on, and neither was the second.

staltz (changes requested):
Looks mostly correct but I have a few suggestions

test/create.test.js
Powersource: on the first one there should be a `group/init` msg

index.js
Powersource: done

test/create.test.js
staltz: Right!

staltz (approved):
All good!

PR #46, opened 2023-01-18
Pin meta feeds to 0.38.1

Powersource:
Tests are broken and i think this fixes them

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
     "ssb-box2": "^4.0.0",
     "ssb-crut": "^4.6.1",
     "ssb-db2": "^6.2.6",
-    "ssb-meta-feeds": "^0.38.1",
+    "ssb-meta-feeds": "0.38.1",
     "ssb-private-group-keys": "^1.1.1",
     "ssb-ref": "^2.16.0",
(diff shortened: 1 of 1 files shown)


staltz:
Do you know why a newer ssb-meta-feeds breaks ssb-tribes2?

Powersource:
tried doing some digging here but didn't come up with much https://github.com/ssbc/ssb-tribes2/pull/44#pullrequestreview-1237351702

staltz:
Seems like it needs more digging, we need these libraries (ssb-db2, ssb-meta-feeds, ssb-tribes2) to be working at their latest versions since we're working on all of them in this project. Otherwise we can't update ssb-meta-feeds in the stack.

Powersource:
is it fine to merge this for now to get tests passing?

staltz (approved):
Please make a "must" issue in ssb-tribes2 to support ssb-meta-feeds >=0.38.2, I think it's something to be fixed as soon as we can, because there is code already using these libraries (e.g. in arj's ssb-browser-demo)

Powersource:
made issue https://github.com/ssbc/ssb-tribes2/issues/48

Issue #41, opened 2022-12-07
addMember + listInvites + acceptInvite


staltz:
The `addMember` API should be renamed to "inviteMember".

The `start()` listener should NOT add the keys to ssb-box2, should NOT perform reindexEncrypted.

A new API `listInvites` that gives a pull-stream source of all the invites (group/add-member) you have received but haven't yet "accepted".

A new `inviteMember` API that allows you to acknowledge a group/add-member, add the keys to ssb-box2, and begin reindexEncrypted.

Powersource:
> The addMember API should be renamed to "inviteMember".

That gives a wrong impression though, since the function *does* cryptographically add the person to the group (gives them the keys)

> A new `inviteMember` API that allows you to acknowledge a group/add-member, add the keys to ssb-box2, and begin reindexEncrypted.

Do you mean "A new `acceptInvite` API" ?

Powersource:
To clarify for nonlinear, this issue is about making the "get added to a group-flow" into:

1. Someone adds (invites) you to a group
2. You see something like "you have been invited to 1 group", and maybe not much more info than that, maybe who invited you. you already have the keys by now, you just have to "accept" by actually using them and actually download the group content
3. you can click accept, then you download the group and can post in it etc

Should this be the flow?

Powersource:
An alternative (current code) is:

1. Someone adds you to a group
2. You automatically download the whole group and can now post in it

staltz:
Okay, we could keep the name `addMember`.

And yes I mean `acceptInvite` API.

> You automatically download the whole group and can now post in it

Without consent, this is pretty bad, because it's a way how users can force other users to download a lot more content, and could be an attack vector.

Powersource:
For Alice to get added to a spam group by Mallory, Alice needs to first be replicating Mallory's invitations feed, which already is an attack vector. Although with a spammed invitations feed I guess it is harder for Mallory to get very far, since everyone replicating her would get spammed. Adding Alice to a specific group would allow Mallory to specifically target her. Unless Alice and Mallory connect directly though, someone else would also have to replicate that feed, and be able to maybe detect the spam.

If Alice doesn't automatically replicate, and instead just sees the invite at first, how much does that help her though? She will have very little info on which to judge if it's spam or not.

staltz:
> If Alice doesn't automatically replicate, and instead just sees the invite at first, how much does that help her though? She will have very little info on which to judge if it's spam or not.

[...]

PR #47, opened 2023-01-18
Require accepting group additions before replicating them

Powersource:
Fixes https://github.com/ssbc/ssb-tribes2/issues/41
Fixes https://github.com/ssbc/ssb-tribes2/issues/17

todos:

* [x] A new API `listInvites` that gives a pull-stream source of all the invites (group/add-member) you have received but haven't yet "accepted".
* [x] A new `acceptInvite` API that allows you to acknowledge a group/add-member, add the keys to ssb-box2, and begin reindexEncrypted.
* [x] Remove `start()`? no it's still needed for creating the additions feed (so triangulation works)
* [x] Update readme
* [x] remove references to `waitUntilMembersOf`?


test/add-member.test.js
```diff
@@ -43,10 +43,15 @@ test('get added to a group', async (t) => {
...
 
-  await alice.tribes2.addMembers(groupId, [bobRoot.id])
+  console.log({ groupId, bobRootId: bobRoot.id })
+  await alice.tribes2.addMembers(groupId, [bobRoot.id]).catch(t.fail)
+  console.log('added bob')
   t.pass('alice added bob to the group')
 
-  await replicate(alice, bob, { waitUntilMembersOf: groupId })
+  console.log('about to replicate')
+  await replicate(alice, bob).catch(t.fail)
```
Powersource, line 52: it gets stuck here, idk why. the only real difference up to this point should be that start() doesn't automatically add invites, so bob doesn't automatically start decrypting alice's group feeds. but bob should still replicate it since i've added `person2.ebt.request(feedId, true)`, right?

test/helpers/replicate.js
```diff
@@ -36,6 +36,7 @@ module.exports = async function replicate(person1, person2, opts = {}) {
     pull.unique(),
     (drain1 = pull.drain((feedId) => {
       person1.ebt.request(feedId, true)
+      person2.ebt.request(feedId, true)
```
Powersource, line 39: remove this file and switch to the new replicate function? not much difference in the end i think but :shrug:

test/helpers/new-replicate.js
```diff
@@ -0,0 +1,136 @@
...
+    }))
+  )
+
+  // Establish a network connection
+  const conn = await p(person1.connect)(person2.getAddress())
+
+  // Wait until both have the same forest
+  //const tree1AtPerson1 = await p(getSimpleTree)(person1, person1Root.id)
+  //const tree2AtPerson2 = await p(getSimpleTree)(person2, person2Root.id)
+  //await retryUntil(async () => {
```
Powersource, line 62: Remove this and `getSimpleTree`?

test/add-member.test.js
Powersource: had a call with andre and we figured out that the problem was that both users had the same feeds fully replicated, the problem was that we were comparing trees too and not only feeds. bob had alice's group feed but he didn't know it was alice's because he couldn't yet decrypt the announce message. so we just removed all the tree equality code. hopefully we don't actually need that?

index.js
```diff
@@ -248,46 +249,87 @@ module.exports = {
...
-          ),
-          pull.asyncMap((msg, cb) => {
-            const groupId = lodashGet(msg, 'value.content.recps[0]')
-
-            ssb.box2.getGroupKeyInfo(groupId, (err, info) => {
-              if (err) {
-                console.error('Error when finding group invite for me:', err)
-                return
-              }
+    function listInvites() {
```
Powersource, line 252: add test for this function specifically. acceptInvite will be tested a lot in other tests

Powersource: or i guess we should also test that things don't work when we don't accept the invite

Powersource: added a test that checks both

test/invites.test.js
```diff
@@ -0,0 +1,79 @@
...
+
+  await alice.tribes2
+    .addMembers(group.id, [bobRoot.id])
+    .catch((err) => t.fail(err))
+  t.pass('alice added bob to the group')
+
+  await replicate(alice, bob).catch(t.fail)
+  t.pass('alice and bob replicate')
+
+  const invites = await pull(bob.tribes2.listInvites(), pull.collectAsPromise())
```
staltz, line 48: Sweet to see collectAsPromise here

test/invites.test.js
```diff
@@ -0,0 +1,79 @@
...
+  t.equal(invites.length, 1, 'bob has 1 invite')
+
+  const invite = invites[0]
+  t.equal(invite.id, group.id, 'correct group id in invite')
+  t.true(invite.secret.equals(group.secret), 'correct secret')
+  t.equal(invite.root, group.root, 'correct root')
+
+  const msgEnc = await p(bob.db.get)(group.root).catch(t.fail)
+  t.true(
+    msgEnc.content && msgEnc.content && !msgEnc.content.type,
```
staltz, line 58:
```suggestion
    typeof msgEnc.content === 'string'
```

Or even better, `t.equals(typeof msgEnc.content, 'string')`

test/invites.test.js
```diff
@@ -0,0 +1,79 @@
...
+
+  const noInvites = await pull(
+    bob.tribes2.listInvites(),
+    pull.collectAsPromise()
+  )
+  t.equal(noInvites.length, 0, 'bob has no invites left')
+
+  const msgUnEnc = await p(bob.db.get)(group.root).catch(t.fail)
+  t.true(
+    msgUnEnc.content && msgUnEnc.content && msgUnEnc.content.type,
```
staltz, line 73:
```suggestion
    typeof msgUnEnc.content === 'object' && msgUnEnc.content.type,
```

index.js
```diff
@@ -248,46 +249,89 @@ module.exports = {
...
+      return pull(
+        pull.values([0]),
+        pull.asyncMap((n, cb) => {
+          ssb.metafeeds.findOrCreate((err, myRoot) => {
+            if (err) throw new Error('Could not find or create my root feed')
```
staltz, line 257:
```suggestion
            if (err) cb(clarify(err, 'Failed to get root metafeed when listing invites'))
```

index.js
```diff
@@ -248,46 +249,89 @@ module.exports = {
...
+
+            ssb.box2.listGroupIds((err, groupIds) => {
+              if (err) throw err
+
+              return cb(
+                null,
+                pull(
```
staltz, line 264: Can you extract this `pull()` chain as a `const source = pull(` and then have a simple one-line `return cb(null, source)`?

index.js
```diff
@@ -248,46 +249,89 @@ module.exports = {
...
-              }
+    function listInvites() {
+      return pull(
+        pull.values([0]),
+        pull.asyncMap((n, cb) => {
+          ssb.metafeeds.findOrCreate((err, myRoot) => {
+            if (err) throw new Error('Could not find or create my root feed')
+
+            ssb.box2.listGroupIds((err, groupIds) => {
+              if (err) throw err
```
staltz, line 260:
```suggestion
              if (err) clarify(err, 'Failed to list group IDs when listing invites'))
```

index.js
```diff
@@ -248,46 +249,89 @@ module.exports = {
...
+                    // it's an addition of us
+                    lodashGet(msg, 'value.content.recps', []).includes(
+                      myRoot.id
+                    )
+                  ),
+                  pull.filter(
+                    (msg) =>
+                      // we haven't already accepted the addition
+                      !groupIds.includes(
+                        lodashGet(msg, 'value.content.recps[0]')
```
staltz, line 279: Just a comment that doesn't require your action in this PR: it would be nice to get rid of lodashGet here and use optional chaining, supported in Node.js 14+ (we are targeting 16+):

`msg?.value?.content?.recps?.[0]`

index.js
```diff
@@ -248,46 +249,89 @@ module.exports = {
...
-            const groupId = lodashGet(msg, 'value.content.recps[0]')
-
-            ssb.box2.getGroupKeyInfo(groupId, (err, info) => {
-              if (err) {
-                console.error('Error when finding group invite for me:', err)
-                return
-              }
+    function listInvites() {
+      return pull(
+        pull.values([0]),
```
staltz, line 254:
```suggestion
        pull.values([0]), // dummy value used to kickstart the stream
```

staltz (changes requested):
Great to see this functioning well. I have just a few comments for code readability

index.js
Powersource: ah lmao that's simpler

Powersource: or ah first i thought you had a solution for avoiding the weird `values([0])` callback stuff. i'll do this now though

staltz (approved):
LGTM!

PR #51, opened 2023-01-24
use clarify-error consistently everywhere

### LINKED ISSUE #31: Use npm module `clarify-error` for all cb errors
See e.g. how ssb-db2 uses it.

staltz:
Closes #31 

The idea is that whenever we bubble up an error, we also attach to it context along the way. ssb-db2 does it like this, so it's nice to follow the same pattern here.


staltz:
@Powersource can haz eyeballz

Powersource (approved):
looks cool!

PR #52, opened 2023-01-24
Fix reindex so it waits for the group key to be added to keyring

arj03:
This uses the updated box2 addGroupInfo to fix a concurrency bug. This also unpins meta-feeds because I believe that should be fixed with https://github.com/ssbc/jitdb/pull/240. So tests might not work right now until that is merged.

### DIFF
--- a/index.js
+++ b/index.js
@@ -379,12 +379,17 @@ module.exports = {
         pull.drain(
           (groupInfo) => {
             foundInvite = true
-            ssb.box2.addGroupInfo(groupInfo.id, {
-              key: groupInfo.secret,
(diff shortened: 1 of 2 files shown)


staltz (approved):
Pre-approving this. It would be nice to merge only when jitdb is updated and the tests pass in CI.

arj03:
Sadly it still fails with latest meta-feeds. I was working on an earlier version of this library, will have a look at a later time why it still fails. At least the fix in this PR is a separate thing and should make the tests more stable so once the tests pass I'll merge this. Thanks for having a look.

Issue #43, opened 2022-12-16
Re-add checking of add-member spec


Powersource:
It's disabled for some reason

https://github.com/ssbc/ssb-tribes2/blob/206138ecabb88f0f807e4d3e2218808d53db79ab/index.js#L21

https://github.com/ssbc/ssb-tribes2/blob/206138ecabb88f0f807e4d3e2218808d53db79ab/index.js#L196-L198

staltz:
It's disabled because the spec checks were failing due to outdated specs.

Powersource:
waiting on https://github.com/ssbc/private-group-spec/pull/23

PR #58, opened 2023-02-10
Re-enable checking of add-member spec

Powersource:
Fixes https://github.com/ssbc/ssb-tribes2/issues/43

### DIFF
--- a/index.js
+++ b/index.js
@@ -467,7 +467,7 @@ module.exports = {
                     return {
                       id: lodashGet(msg, 'value.content.recps[0]'),
                       secret: Buffer.from(
-                        lodashGet(msg, 'value.content.secret'),
+                        lodashGet(msg, 'value.content.groupKey'),
                         'base64'
                       ),
                       root: lodashGet(msg, 'value.content.root'),
--- a/index.js
+++ b/index.js
@@ -310,7 +310,7 @@ module.exports = {
       return pull(ssb.box2.listGroupIds({ live: !!opts.live }), paraMap(get, 4))
     }
 
-    function getRootIdOfMsgAuthor(groupRootMsgId, cb) {
+    function getRootFeedIdFromMsgId(groupRootMsgId, cb) {
       ssb.db.get(groupRootMsgId, (err, rootMsg) => {
         // prettier-ignore
         if (err) return cb(clarify(err, "couldn't get root msg for finding root feed"))
@@ -342,7 +342,7 @@ module.exports = {
         // prettier-ignore
         if (err) return cb(clarify(err, `Failed to get group details when adding members`))
 
-        getRootIdOfMsgAuthor(root, (err, rootAuthorId) => {
+        getRootFeedIdFromMsgId(root, (err, rootAuthorId) => {
           // prettier-ignore
           if (err) return cb(clarify(err, "couldn't get root id of author of root msg"))
 
--- a/index.js
+++ b/index.js
@@ -335,19 +335,7 @@ module.exports = {
                   // prettier-ignore
                   if (err) return cb(clarify(err, "couldn't find parent of shard feed"))
 
-                  console.log('v1Feed', v1Feed)
-
-                  ssb.metafeeds.advanced.findById(
-                    v1Feed.parent,
-                    (err, rootFeed) => {
-                      // prettier-ignore
-                      if (err) return cb(clarify(err, "couldn't find root feed from v1 feed"))
-
-                      console.log('supposed root feed', rootFeed)
-
-                      return cb(null, rootFeed.id)
-                    }
-                  )
+                  return cb(null, v1Feed.parent)
                 }
               )
             }
@@ -378,12 +366,11 @@ module.exports = {
           // prettier-ignore
           if (err) return cb(clarify(err, "couldn't get root id of author of root msg"))
 
-          console.log('rootAuthorId', rootAuthorId)
-
           const content = {
             type: 'group/add-member',
             version: 'v2',
             groupKey: secret.toString('base64'),
+            root,
             creator: rootAuthorId,
             tangles: {
               members: {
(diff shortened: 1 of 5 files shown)


index.js
```diff
@@ -310,6 +314,48 @@ module.exports = {
...
+
+                  ssb.metafeeds.advanced.findById(
+                    v1Feed.parent,
+                    (err, rootFeed) => {
+                      // prettier-ignore
+                      if (err) return cb(clarify(err, "couldn't find root feed from v1 feed"))
+
+                      console.log('supposed root feed', rootFeed)
+
+                      return cb(null, rootFeed.id)
```
Powersource, line 348: Weird bug here https://github.com/ssbc/ssb-meta-feeds/issues/115

Powersource: reproduced https://github.com/ssbc/ssb-meta-feeds/pull/116

Powersource: lol realized i already have what i need at `v1Feed.parent`. still nice bug find tho

index.js
```diff
@@ -310,6 +314,36 @@ module.exports = {
...
+          // prettier-ignore
+          if (err) return cb(clarify(err, "couldn't get root msg for finding root feed"))
+
+          ssb.metafeeds.advanced.findById(
+            groupFeed.parent,
+            (err, shardFeed) => {
+              // prettier-ignore
+              if (err) return cb(clarify(err, "couldn't find parent of group feed"))
+
+              ssb.metafeeds.advanced.findById(
```
mixmix, line 332: @Powersource I think we want a method on meta feeds - `findRootFeedId(subfeedId,cb)

Powersource: @mixmix made a PR https://github.com/ssbc/ssb-meta-feeds/pull/119

package.json
```diff
@@ -39,14 +39,14 @@
...
-    "private-group-spec": "^2.0.0",
+    "private-group-spec": "^3.0.0",
     "pull-paramap": "^1.2.2",
     "pull-stream": "^3.7.0",
     "ssb-bfe": "^3.7.0",
     "ssb-box2": "^5.0.0",
     "ssb-crut": "^4.6.1",
     "ssb-db2": "^6.3.0",
-    "ssb-meta-feeds": "^0.38.2",
+    "ssb-meta-feeds": "github:ssbc/ssb-meta-feeds#find-root-feed-id",
```
Powersource, line 49: https://github.com/ssbc/ssb-meta-feeds/pull/119

index.js
```diff
@@ -310,6 +310,20 @@ module.exports = {
       return pull(ssb.box2.listGroupIds({ live: !!opts.live }), paraMap(get, 4))
     }
 
+    function getRootIdOfMsgAuthor(groupRootMsgId, cb) {
```
Powersource, line 313: move into metafeed helper file

Powersource: ah i guess we don't have that file yet https://github.com/ssbc/ssb-tribes2/pull/59

index.js
```diff
@@ -328,41 +342,46 @@ module.exports = {
...
-        // TODO: this should accept bendybutt-v1 feed IDs
-        // if (!addMemberSpec(content))
-        //   return cb(new Error(addMemberSpec.errorsString))
+          if (opts.text) content.text = opts.text
 
-        findOrCreateAdditionsFeed((err, additionsFeed) => {
-          // prettier-ignore
-          if (err) return cb(clarify(err, 'Failed to find or create additions feed when adding members'))
+          if (!addMemberSpec(content))
+            return cb(new Error(addMemberSpec.errorsString))
```
Powersource, line 373: Could we move this down to between `addGroupTangle` and `publishAndPrune`, as well as remove the group tangle above, to make the checking more true to reality?

package.json
mixmix: Replace before merge

index.js
mixmix, line 313:
```suggestion
    function getRootFeedIdFromMsg(groupRootMsgId, cb) {
```

I didn't quite understand your function name. I think it should be something like this.
I think this surfaces that we have a need for a new name. Author is kinda feed-specific in current ssb.
I wonder if it's `deviceId` - the feedId of the root meta-feed for a device. God.
`rootAuthor`

`getRootAuthorByMsgId`

index.js
```diff
@@ -328,41 +342,46 @@ module.exports = {
...
+        getRootIdOfMsgAuthor(root, (err, rootAuthorId) => {
+          // prettier-ignore
+          if (err) return cb(clarify(err, "couldn't get root id of author of root msg"))
+
+          const content = {
+            type: 'group/add-member',
+            version: 'v2',
+            groupKey: secret.toString('base64'),
+            root,
+            creator: rootAuthorId,
```
mixmix, line 354: I might have missed this (`creator`) being added to the spec? I dislike that it's like author. Why do we need it?

> creator is the root metafeed id of the creator of the group 

Nope, I was involved! I think we should call it `groupCreatorRootFeedId` or `initAuthorRootrId` .. if it's not too late :D

index.js
```diff
@@ -328,41 +342,46 @@ module.exports = {
...
 
-        // TODO: this should accept bendybutt-v1 feed IDs
-        // if (!addMemberSpec(content))
-        //   return cb(new Error(addMemberSpec.errorsString))
+          if (opts.text) content.text = opts.text
 
-        findOrCreateAdditionsFeed((err, additionsFeed) => {
-          // prettier-ignore
-          if (err) return cb(clarify(err, 'Failed to find or create additions feed when adding members'))
+          if (!addMemberSpec(content))
```
mixmix, line 372: prefer
```suggestion
          if (!isAddMember(content))
```

index.js
mixmix: Sounds like I a good idea

mixmix (commented):
Looks ready apart from waiting for that module to be installed when it's published

index.js
Powersource: oh yeah that's a better function name

index.js
Powersource: Not too late! But this PR just adds the checking against the spec. Changing the field name would IMO be a different issue/pr.

PR #59, opened 2023-02-12
split out meta feed shuffling into helpers

mixmix:
Was reading over code, and found the index.js file overwhelming. Split out functions that were meta-feed related into a single file, so the API parts of the code were clearer/ easier to find.

I also added a function `drainOne` because the "take + drain" pattern looks dangerous to me - if someone changes that `take(1)` at any point then you are gonna get multiple calls of `cb`, which would create weird / bad bugs

### DIFF
--- a/index.js
+++ b/index.js
@@ -21,6 +21,7 @@ const AddGroupTangle = require('./lib/add-group-tangle')
 const MetaFeedHelpers = require('./lib/meta-feed-helpers')
 const buildGroupId = require('./lib/build-group-id')
 const publishAndPrune = require('./lib/prune-publish')
+const drainOne = require('./lib/drain-one')
 
 module.exports = {
   name: 'tribes2',
@@ -312,8 +313,4 @@ module.exports = {
       start,
     }
   },
-}
-
-function drainOne(handleErr, handleRes) {
-  return pull(pull.take(1), pull.drain(handleErr, handleRes))
 }
--- a/index.js
+++ b/index.js
@@ -10,24 +10,16 @@ const clarify = require('clarify-error')
 const {
   where,
   and,
-  count,
   isDecrypted,
   type,
   live,
-  author,
-  toCallback,
   toPullStream,
 } = require('ssb-db2/operators')
-const {
-  keySchemes,
-  validator: {
-    group: { init: initSpec, addMember: addMemberSpec },
-  },
-} = require('private-group-spec')
-const { SecretKey } = require('ssb-private-group-keys')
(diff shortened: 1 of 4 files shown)


lib/meta-feed-helpers.js
```diff
@@ -0,0 +1,242 @@
...
+
+            if (count === 0) {
+              // we're only interested in empty feeds
+              return cb(null, feed)
+            } else {
+              return cb()
+            }
+          })
+        )
+      }, 4),
```
mixmix, line 61: I also added a `width` to this paramap (without width, paraMap behaves like `pull.async` , i.e. no parallel boost)

lib/meta-feed-helpers.js
```diff
@@ -0,0 +1,246 @@
...
+      pull.filter(Boolean),
```
staltz, line 66: What happened to the `pull.unique('id')`?

lib/meta-feed-helpers.js
```diff
@@ -0,0 +1,246 @@
...
+    secretKeyFromPurpose,
+
+    findOrCreateAdditionsFeed,
+    findOrCreateGroupFeed,
+    findOrCreateGroupWithoutMembers
+  }
+}
+
+function drainOne(handleErr, handleRes) {
+  return pull(pull.take(1), pull.drain(handleErr, handleRes))
```
Powersource, line 245: wrong arg names https://pull-stream.github.io/#drain

index.js
```diff
@@ -526,3 +313,7 @@ module.exports = {
     }
   },
 }
+
+function drainOne(handleErr, handleRes) {
```
Powersource, line 317: duplicate

lib/meta-feed-helpers.js
mixmix: removed because it was un-needed - there is no need for unique if you're only taking 1!

index.js
mixmix: extracted

mixmix:
Ready for review again @Powersource

lib/prune-publish.js
```diff
@@ -16,6 +16,9 @@ module.exports = function prunePublish(ssb, content, keys, cb) {
 
         return prunePublish(ssb, content, keys, cb)
       } else {
+        console.log(content)
+        console.log(keys)
+        console.log(err)
```
Powersource, line 21: remove

Powersource (approved):

Powersource:
replaced by https://github.com/ssbc/ssb-tribes2/pull/66

PR #60, opened 2023-02-13
Unpin ssb-meta-feeds

### LINKED ISSUE #48: Unpin meta-feeds version
context https://github.com/ssbc/ssb-tribes2/pull/46

arj03:
... and add workaround in replicate to better simulate real EBT.

This should now work. A bit hacky. Might need https://github.com/ssbc/ssb-meta-feeds/pull/118 to work. This is working locally with that patch.

Should fix #48

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
     "ssb-box2": "^5.0.0",
     "ssb-crut": "^4.6.1",
     "ssb-db2": "^6.3.0",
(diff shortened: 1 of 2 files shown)


staltz (approved):
Dang, pretty deep. LGTM

Powersource (approved):
Feels wrong to add a hack/rely on a complex module that's supposed to be optional (repl. scheduler) but i guess this needs to be updated

PR #61, opened 2023-02-15
Check against content spec when publishing

Powersource:
Fixes https://github.com/ssbc/ssb-tribes2/issues/27

### DIFF
--- a/index.js
+++ b/index.js
@@ -21,7 +21,7 @@ const {
 const {
   keySchemes,
   validator: {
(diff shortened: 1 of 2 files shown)


mixmix (approved):
I think "spec" is a bit generic a term to be throwing around.
The pattern I've used in the rest of ssb is `isValidPost` or `isValidContent` etc.

I personally like the `isX` pattern as a signal that it returns type `Boolean`

PR #62, opened 2023-02-16
Add groupkey in init msg

Powersource:
for https://github.com/ssbc/ssb-tribes2/issues/40

### DIFF
--- a/index.js
+++ b/index.js
@@ -143,29 +143,30 @@ module.exports = {
       })
     }
 
-    function secretKeyFromPurpose(purpose) {
-      return new SecretKey(Buffer.from(purpose, 'base64'))
+    function secretKeyFromString(string) {
+      return new SecretKey(Buffer.from(string, 'base64'))
     }
 
     function createGroupWithoutMembers(myRoot, cb) {
-      const content = {
-        type: 'group/init',
-        tangles: {
-          group: { root: null, previous: null },
-        },
-      }
-      if (!initSpec(content)) return cb(new Error(initSpec.errorsString))
-
       findOrCreateGroupFeed(null, function gotGroupFeed(err, groupFeed) {
         // prettier-ignore
         if (err) return cb(clarify(err, 'Failed to find or create group feed when creating a group'))
 
-        const secret = secretKeyFromPurpose(groupFeed.purpose)
+        const secret = secretKeyFromString(groupFeed.purpose)
 
         const recps = [
           { key: secret.toBuffer(), scheme: keySchemes.private_group },
           myRoot.id,
         ]
+
+        const content = {
+          type: 'group/init',
+          groupKey: secret.toString('base64'),
+          tangles: {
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -39,7 +39,7 @@
     "fast-deep-equal": "^3.1.3",
     "lodash.get": "^4.4.2",
     "lodash.set": "^4.3.2",
-    "private-group-spec": "^3.0.0",
+    "private-group-spec": "github:ssbc/private-group-spec#groupkey-in-init",
```
Powersource, line 42: https://github.com/ssbc/private-group-spec/pull/25

index.js
```diff
@@ -143,30 +143,31 @@ module.exports = {
...
-        },
-      }
-      if (!initSpec(content)) return cb(new Error(initSpec.errorsString))
-
       findOrCreateGroupFeed(null, function gotGroupFeed(err, groupFeed) {
         // prettier-ignore
         if (err) return cb(clarify(err, 'Failed to find or create group feed when creating a group'))
 
-        const secret = secretKeyFromPurpose(groupFeed.purpose)
+        const secret = secretKeyFromString(groupFeed.purpose)
```
staltz, line 155: Not a big deal, but a small suggestion:

```suggestion
        const groupKey = secretKeyFromString(groupFeed.purpose)
```

We seem to be going back and forth with this naming, "secret" or "group key", but we could try to be at least internally consistent (all namings of this thing in ssb-tribes2 are the same).

index.js
```diff
@@ -257,37 +258,37 @@ module.exports = {
...
 
-            return cb(null, data)
-          })
+        const data = {
+          id: buildGroupId({
+            groupInitMsg,
+            groupKey: secret.toBuffer(),
+          }),
+          secret: secret.toBuffer(),
```
staltz, line 274: Similarly, could name this `groupKey`

index.js
```diff
@@ -257,37 +258,37 @@ module.exports = {
...
+          root: fromMessageSigil(groupInitMsg.key),
+          subfeed: groupFeed.keys,
         }
-      )
+
+        ssb.box2.addGroupInfo(data.id, {
+          key: data.secret,
```
staltz, line 280: Case in point: the key is the secret is the key is the secret.

staltz:
Tests need to be updated, they are failing just because the new field `groupKey` was unexpected.

index.js
Powersource: Problem here is the spec calls it groupKey and the tribes2 api says secret. I think the proper way for fixing is a separate issue/pr.

Powersource: this pr is just for conforming to the spec

staltz: Okay!

package.json
staltz: After merging that one, and publishing, then we can merge this tribes2 PR

staltz (approved):
LGTM!

PR #63, opened 2023-02-17
Start slow

mixmix:
experiment in speeding up tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
     "ssb-box2": "^5.0.0",
     "ssb-crut": "^4.6.1",
     "ssb-db2": "^6.3.0",
-    "ssb-meta-feeds": "^0.38.2",
+    "ssb-meta-feeds": "github:ssbc/ssb-meta-feeds#skipMainFeed",
     "ssb-private-group-keys": "^1.1.1",
     "ssb-ref": "^2.16.0",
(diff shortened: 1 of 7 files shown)


mixmix:
Sigh, not sure why this doesn't work

staltz:
I'm not sure it's worth spending time on this, but a quick inspection shows that most of the time is spent idle. There are a bunch of 200ms or 1000ms waits in between. This is the profiler for the add-member test file:

Screenshot from 2023-02-17 12-00-15@2x

mixmix:
Can you show me how to generate profiler things like that? Are you running an inspector thing to chrome?
There's a bunch of waits between the feeds being added in replication. like 200ms between each... presumably some race condition in `ssb-meta-feeds` ...
But I also notice that startup is really slow with meta feeds (3 seconds!). This is likely the first time cost only, not sure.

staltz:
`node --inspect-brk test/the-file.js` and then open Chrome, go to the address `chrome://inspect` and click the link that it detected on that page.

Notice that from https://github.com/ssbc/ssb-meta-feeds/pull/121#issuecomment-1434428800 I think these 200ms waits are from using lodash.debounce in ssb-db2 and async-append-only-log. Some of which are customizable with a config, others aren't.

Also, beware of yak shaving. I think your time could be more useful elsewhere.

Issue #73, opened 2023-03-16
Update group/init schema


Powersource:
Or do that + create a new schema for new epochs as well?

we need
* epoch tangle init (and updates)
* group tangle should also be able to be updated
* members tangle should always just be init (it's updated in add-member and exclude)

Issue #68, opened 2023-03-13
Add spec checking for exclusion messages


Powersource:
Followup from https://github.com/ssbc/ssb-tribes2/pull/65

PR #65, opened 2023-02-22
Basic member exclusion flow

Powersource:
Starting to implement https://github.com/ssbc/ssb-group-exclusion-spec

* [x] Publish init msg on new epoch.
* [x] Test that messages are published like they should
* [x] Message sharing new key to existing members should be add-member message
* [x] Message announcing exclude should follow exclude spec
* [x] Init epoch tangle on root. Moved here https://github.com/ssbc/ssb-tribes2/issues/73
* [x] Make sure we ourselves can post on the new key
* [x] `clarify` errors
* [x] Update readme
* [x] `create` says "NOTE: If `create` finds an empty (i.e. seemingly unused) group feed, it will start using that feed instead of creating a new one.". I think we might run into that situation the way the new `feedKeys` options work.
* [x] check that `create`'s recovery search for "add-member" messages works

PR #74, opened 2023-03-21
Add spec checking for epoch inits and exclude messages

Powersource:
Fixes https://github.com/ssbc/ssb-tribes2/issues/73
Fixes https://github.com/ssbc/ssb-tribes2/issues/68

Based on https://github.com/ssbc/ssb-tribes2/pull/65


staltz (commented):
Tastes good, but let's hear from Mix

index.js
```diff
@@ -229,7 +234,7 @@ module.exports = {
                       }
                       const newTangleOpts = {
                         tangles: ['epoch'],
-                        isValid: () => true,
+                        isValid: isInit,
```
mixmix, line 237: as I said in the other PR, keen to see `isInitNewEpoch` or something to make sure this is really tight

lib/meta-feed-helpers.js
```diff
@@ -184,7 +185,9 @@ module.exports = (ssb) => {
           toPullStream()
         ),
         // find only init for epoch zero
-        pull.filter(rootMsg => rootMsg.value.content?.tangles?.group?.root === null),
+        pull.filter(
+          (rootMsg) => rootMsg.value.content?.tangles?.group?.root === null
```
mixmix, line 189: with a specific `isInitZero` or `isInitRoot` we could replace this line with a very tight check.
I would be keen to see that because just checking the type and tangle isn't safe enough to lean on it for other fields.

mixmix (approved):
Looks good. As I've said I think we should split out those schema to pin this down tight

lib/meta-feed-helpers.js
Powersource: idk we don't really use the other fields here, we just want the msgKey I think and the schema doesn't help verify that

mixmix: ok. the other thing we could do (maybe make it a TODO) is the put this filter in the query.
i.e. that it is a group tangle root. It would be much faster to not have to parse the json of each `group/init` before checking... but this is just an optimisation

mixmix (approved):

mixmix:
re-running tests because the `retryUntil` timed out :cry:

mixmix:
wuh woh

image

Issue #72, opened 2023-03-16
Test staying in a group when someone else removes someone


Powersource:
Related to https://github.com/ssbc/ssb-tribes2/issues/71

PR #78, opened 2023-03-24
People not removed in an exclude should be able to post on the new epoch

Powersource:
Fixes https://github.com/ssbc/ssb-tribes2/issues/72


test/exclude-members.test.js
```diff
@@ -154,3 +155,122 @@ test('add and remove a person, post on the new feed', async (t) => {
...
+      text: 'second post',
+      recps: [groupId],
+    })
+    .catch(t.fail)
+
+  t.notEquals(
+    secondFeedId,
+    firstFeedId,
+    'feed for second publish is different to first publish'
+  )
```
Powersource, line 250: This fails.

Checked replication and carol seems to catch up to the latest message on the feed that the re-addition is published to by alice, but only alice seems to find the re-addition in start(). Does that mean that carol doesn't manage to decrypt the addition? Still happens with 10 seconds timeout before carol publishes.

Powersource: in box2/monitor.js we need to change it to something like

```js
      //pull.filter(isForeignLeafFeedBranch),
      pull.asyncMap((leafBranch, cb) => {
        if (!rootKeys) return cb()
        if (isForeignLeafFeedBranch(leafBranch)) {
          const theirRoot = leafBranch[0]
          const theirLeaf = leafBranch[3]
          ssb.metafeeds.getTree(rootKeys.id, (err, myRoot) => {
```
with an else where we do the reverse case

Powersource: andré described the problem like:

The problem was that: Carol tries to create a triangle with Alice as soon as Carol discover's Alice's epoch 0 feed, BUT at that point Carol doesn't yet have an epoch 0 feed, so creating triangle fails, and Carol never tries it again.

However, Carol SHOULD create a triangle with Alice's epoch 0 feed as soon as Carol creates her own epoch 0 feed

Powersource: We're putting this bug in the "fridge" for now by putting all add-member msgs on the additions feed, something we were thinking about anyway. TODO make issue etc

Powersource: Andre got a task to update the exclusion spec. Should make issue for this repo

Powersource: Rewrote this issue to be just about moving stuff to the additions feed. Will work on that now #79

test/exclude-members.test.js
```diff
@@ -154,3 +155,117 @@ test('add and remove a person, post on the new feed', async (t) => {
...
+    .publish({
+      type: 'test',
+      text: 'first post',
+      recps: [groupId],
+    })
+    .catch((err) => t.error(err, 'carol failed to publish on first feed'))
+
+  await alice.tribes2
+    .excludeMembers(groupId, [bobRoot.id])
+    .catch((err) => t.error(err, 'remove member fail'))
```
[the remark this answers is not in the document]
Powersource: Unresolving this, we should test this here once #79 is fixed

test/exclude-members.test.js
```diff
@@ -174,3 +175,133 @@ test('add and remove a person, post on the new feed', async (t) => {
...
+    firstFeedId,
+    'feed for second publish is different to first publish'
+  )
+
+  const { writeKey: writeKey2 } = await carol.tribes2.get(groupId)
+
+  const branches = await pull(
+    carol.metafeeds.branchStream({ root: carolRoot, old: true, live: false }),
+    pull.collectAsPromise()
+  )
```
Powersource, line 287: after basing on #85 this

```js
  const branches = await pull(
    carol.metafeeds.branchStream({ root: carolRoot, old: true, live: false }),
    pull.collectAsPromise()
  )

  console.log('carol branches', branches)
```

logs 

```js
carol branches [
  [
    FeedDetails {
      id: [FeedDetails],
      parent: null,
      purpose: 'root',
      feedFormat: 'bendybutt-v1',
      seed: null,
      keys: null,
      recps: null,
      metadata: {},
      tombstoned: false,
      tombstoneReason: null
    }
  ]
]
```

even though carol has successfully posted twice to the group, including on different feeds

test/exclude-members.test.js
Powersource: seems like this might work after merging #85 into this

test/exclude-members.test.js
Powersource: realized that carolRoot is an object and not the root id. strange that it returns anything at all

Powersource: made related pr https://github.com/ssbc/ssb-meta-feeds/pull/122

test/exclude-members.test.js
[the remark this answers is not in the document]
Powersource: booya works

index.js
```diff
@@ -231,10 +231,11 @@ module.exports = {
                     // prettier-ignore
                     if (err) return cb(clarify(err, "Couldn't post init msg on new epoch when excluding members"))
 
-                    addMembers(groupId, remainingMembers, {}, (err) => {
+                    addMembers(groupId, remainingMembers, {}, (err, reMsg) => {
                       // prettier-ignore
                       if (err) return cb(clarify(err, "Couldn't re-add remaining members when excluding members"))
-                      return cb()
+                      //TODO: we probably don't want to return this exactly
+                      return cb(null, { excludeMsg, reAddMsg: reMsg })
```
Powersource, line 238: todo remove probably

index.js
```diff
@@ -407,6 +408,78 @@ module.exports = {
         if (err) return cb(clarify(err, 'Error finding or creating additions feed when starting ssb-tribes2'))
         return cb()
       })
+
+      // look for new epochs that we're added to
+      ssb.metafeeds.findOrCreate((err, myRoot) => {
+        // prettier-ignore
+        if (err) return cb(clarify(err, 'todo'))
```
Powersource, line 415: clarify errs

index.js
```diff
@@ -407,6 +408,78 @@ module.exports = {
...
+
+        pull(
+          ssb.db.query(
+            // TODO: does this output new stuff if we accept an invite to an old epoch and then find additions to newer epochs?
+            where(and(isDecrypted('box2'), type('group/add-member'))),
+            live({ old: true }),
+            toPullStream()
```
Powersource, line 422: make issue about creating test case for this

index.js
```diff
@@ -407,6 +408,78 @@ module.exports = {
...
+          ),
+          // groups/epochs we're added to
+          pull.filter((msg) => {
+            console.log('found an addition', myRoot.id, msg.value.content.recps)
```
Powersource, line 426: remove all the console logs

index.js
```diff
@@ -407,6 +408,78 @@ module.exports = {
...
+              ssb.box2.addGroupInfo(groupId, { key: newKey }, (err) => {
+                // prettier-ignore
+                if (err) return cb(clarify(err, 'todo'))
+
+                const newKeyPick = {
+                  key: newKey,
+                  scheme: keySchemes.private_group,
+                }
+                // TODO: naively guessing that this is the latest key for now
+                ssb.box2.pickGroupWriteKey(groupId, newKeyPick, (err) => {
```
Powersource, line 463: Make issue about making this logic more advanced

test/exclude-members.test.js
```diff
@@ -174,3 +175,142 @@ test('add and remove a person, post on the new feed', async (t) => {
...
+    .then((res) => {
+      t.pass('alice excluded bob')
+      return res
+    })
+    .catch((err) => t.error(err, 'remove member fail'))
+
+  await replicate(alice, carol).catch(t.error)
+
+  // TODO: maybe remove?
+  await p(setTimeout)(10000)
```
Powersource, line 257: remove or shorten

index.js
Powersource: removed

index.js
Powersource: https://github.com/ssbc/ssb-tribes2/issues/86

index.js
Powersource: https://github.com/ssbc/ssb-tribes2/issues/87

index.js
```diff
@@ -407,6 +407,72 @@ module.exports = {
...
+                if (err) return cb(clarify(err, "Error getting groups we're already in when looking for new epochs"))
+
+                if (groupIds.includes(msg.value?.content?.recps?.[0])) {
+                  return cb(null, msg)
+                } else {
+                  return cb()
+                }
+              })
+            )
+          }),
```
mixmix, line 442: `paraMap(fn) === pull.asyncMap(fn)`
You need to add a "width" for it to be running in parallel e.g. run at most 4 at once:

```suggestion
          }, 4),
```

index.js
```diff
@@ -407,6 +407,72 @@ module.exports = {
...
+                // prettier-ignore
+                if (err) return cb(clarify(err, "Error getting groups we're already in when looking for new epochs"))
+
+                if (groupIds.includes(msg.value?.content?.recps?.[0])) {
+                  return cb(null, msg)
+                } else {
+                  return cb()
+                }
+              })
+            )
```
mixmix, line 441: we can move faster by comparing as we go and stopping as soon as we find a match

```suggestion
            const groupId = msg.value?.content?.recps?.[0]
            if (!groupId) return cb(null, null)

            pull(
              ssb.box2.listGroupIds(),
              pull.filter(id => id === groupId),
              pull.take(1),
              pull.collect((err, groupIds) => {
                // prettier-ignore
                if (err) return cb(clarify(err, "Error getting groups we're already in when looking for new epochs"))

                if (groupIds.length) return cb(null, msg)
                else cb(null, null)
              })
            )
```

index.js
```diff
@@ -407,6 +407,72 @@ module.exports = {
...
+          pull.drain(
+            (msg) => {
+              const groupId = msg.value?.content?.recps?.[0]
+
+              const newKey = Buffer.from(msg.value?.content?.groupKey, 'base64')
+              ssb.box2.addGroupInfo(groupId, { key: newKey }, (err) => {
+                // prettier-ignore
+                if (err) return cb(clarify(err, 'Error adding new epoch key that we found'))
+
+                const newKeyPick = {
```
mixmix, line 453: I suspect we may have to record the epoch root for each key. I don't think it's sufficient to just swap in the latest key as the  correct writeKey.
Ah you have a TODO about it being naive, nice

BTW I really like your PR style of "here is a bit" and clearly leaving TODOs which are for the next round. When I worked on tribes 1 I loaded everything up in my head and was like "it's all got to work" :facepalm:

index.js
```diff
@@ -407,6 +407,72 @@ module.exports = {
...
+        pull(
+          ssb.db.query(
+            // TODO: does this output new stuff if we accept an invite to an old epoch and then find additions to newer epochs?
+            where(and(isDecrypted('box2'), type('group/add-member'))),
+            live({ old: true }),
+            toPullStream()
+          ),
+          // groups/epochs we're added to
+          pull.filter((msg) => {
+            return msg.value?.content?.recps?.includes(myRoot.id)
```
mixmix, line 425:
```suggestion
            return isGroupAdd(msg) && msg.value.content.recps.includes(myRoot.id)
```

mixmix (changes requested):
one :fire: (must change)
minor performance suggestions

Good clear test <3

index.js
```diff
@@ -407,6 +407,72 @@ module.exports = {
...
+                } else {
+                  return cb()
+                }
+              })
+            )
+          }),
+          pull.filter(Boolean),
+          pull.drain(
+            (msg) => {
+              const groupId = msg.value?.content?.recps?.[0]
```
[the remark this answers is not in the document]

index.js

index.js
Powersource: listGroupIds is basically sync, it's just a stream to allow for its live opt

index.js
Powersource: checked against the spec in a filter before this so removed the ?s

mixmix (approved):

PR #80, opened 2023-03-30
speed up request step for replicate

mixmix:
This PR modifies replicate helper in tests to remove redundent 200ms waits (and redundent streaming of meta feeds if not needed)

Test suite times on my desktop:

BEFORE:  264.968 s
AFTER:   252.572 s

FML


mixmix:
I think the gains are really in tests where you replicate multiple times.... I feel like it was making a big different in my branch, but maybe not

Powersource (changes requested):
Tests failing.

mixmix:
no I was hoping it had made more difference. I think lets close this for now

PR #83, opened 2023-04-04
lib/epoch - getEpochs

mixmix:
NOTE: to be merged into tangles-refactor  (or after that)

This is a "tiny" PR which just lets you get all the epochs for a particular group
I am not sure if this is to be a public/advertised API so may do that later.

This PR will be followed by others which add more detailed queries like "who are the members of an epoch", and "which is the preferred epoch" and "are all the members correct for this epoch?"  more low level tools for doing the heavy lifting


test/lib/epochs.test.js
```diff
@@ -0,0 +1,129 @@
...
+  }
+
+  await run(
+    'start tribes',
+    Promise.all(peers.map(peer => p(peer.tribes2.start)())),
+  )
+
+  const group = await run(
+    'alice creates a group',
+    p(alice.tribes2.create)({})
```
Powersource, line 56: :fire: avoid using promisify on tribes2 functions. it breaks stuff sometimes since they self-promisify

test/lib/epochs.test.js
```diff
@@ -0,0 +1,129 @@
...
+    epochs,
+    [
+      {
+        key: group.root,
+        previous: null,
+        epochKey: group.writeKey.key,
+        author: epochs[0].author // CHEAT
+      },
+      {
+        key: epochs[1].key, // CHEAT ...get this from excludeMembers?
```
Powersource, line 119: haven't decided what, if anything, to return from excludeMembers yet. do you have ideas?

lib/epochs.js
```diff
@@ -0,0 +1,118 @@
...
+    type: 'object',
+    properties: ['author', 'epochKey'],
+    required: ['author', 'epochKey'],
+    author: {
+      type: 'string',
+      pattern: feedPattern
+    },
+    epochKey: {
+      type: 'string',
+      pattern: '[a-zA-Z0-9/\\+]{43}=' // 32 bytes in base64
```
Powersource, line 42: this and the above regex feel a bit like magic numbers, that we've written in several places. is there a way to import at least some of this from somewhere else? for the author ssb-uri2 might help?

lib/epochs.js
```diff
@@ -0,0 +1,118 @@
...
+    }
+  }
+})
+strategy.mapToPure = T => T || strategy.identity() // function @tangle/reduce needs
```
Powersource, line 46: this is sort of the black box magic that i mentioned. i have no idea what this means. maybe that's ok though

lib/epochs.js
Powersource: is it simpler maybe to do like you've suggested before and run the normal validate functions directly  on the messages around line 87?

Powersource (changes requested):
mostly discussion starters but removing the promisifies would be nice

test/lib/epochs.test.js
```diff
@@ -0,0 +1,129 @@
...
+      isTest = true,
+      timer = false,
+      logError = false
+    } = opts
+
+    if (timer) console.time('> ' + label)
+    return promise
+      .then(res => {
+        if (isTest) t.pass(label)
+        else console.log(' ', label)
```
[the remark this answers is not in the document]
mixmix: Sure, I can remove the console.log

test/lib/epochs.test.js
mixmix: GOOD TO KNOW! I will undo this

test/lib/epochs.test.js
```diff
@@ -0,0 +1,129 @@
...
+    alice.tribes2.getEpochs(group.id)
+  )
+  t.deepEqual(
+    epochs,
+    [{
+      key: group.root,
+      previous: null,
+      epochKey: group.writeKey.key,
+      author: epochs[0].author // CHEAT
+      // TODO where can we get the author feed from... do we need it?
```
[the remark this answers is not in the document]
mixmix: Thanks, that's a good point.... but are we expecting the root author or the .... oh we can find the epoch using the key... I forgot. DUH

test/lib/epochs.test.js
mixmix: Yeah I thought maybe the same pattern that comes back from `create` as it's kinda similar.... except that `info.subfeed` is really anomolous - it's raw fucking ssb-keys, and ... they're not in the URI form id..

index.js
```diff
@@ -436,6 +439,8 @@ module.exports = {
       listInvites,
       acceptInvite,
       start,
+
+      getEpochs,
```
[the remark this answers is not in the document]
mixmix: Yeah I don't know what needs to be public. Could be this ends up not being public?
LATER

lib/epochs.js
mixmix: uhhh .... this builds the validate functions are runs them for you in `reduce`
But yes I'll try and find a more canonical source for these pattern strings.

lib/epochs.js
mixmix: yeah ... it's because I'm using low level tools.
I will perhaps leave a longer comment about what `@tangle/reduce` is bringing to the party

lib/epochs.js
```diff
@@ -0,0 +1,118 @@
...
+    if (err) return cb(err)
+
+    ssb.db.get(info.root, (err, rootVal) => {
+      if (err) return cb(err)
+      // TODO check is a valid groupInitRoot
+      const root = { key: info.root, value: rootVal }
+
+      getTangleUpdates(ssb, 'epoch', info.root, (err, updates) => {
+        if (err) return cb(err)
+        // TODO filter updates usiny groupInitEpoch
```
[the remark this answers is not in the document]
mixmix: once  those validators are published I'll use them here

test/lib/epochs.test.js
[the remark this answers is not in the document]
mixmix: yeah I've made it the id of the root metafeed

lib/epochs.js
mixmix: `ssb-uri2` does not expose its patterns / regexp. Annoying, and outside of scope of this. I'll leave a note in the code

mixmix: ok I built a module. This is your fault
https://github.com/ssbc/butt64

index.js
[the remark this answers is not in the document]
mixmix: removed and just testing the lib export directly

lib/epochs.js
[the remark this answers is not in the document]
mixmix: done

lib/epochs.js
Powersource: wow cool!

Powersource (approved):

Powersource:
just fix the test :)

Powersource:
was gonna say fix the test but saw you commented that it's failing because exclusion is still wip. i guess fair enough. feels a bit weird maybe to merge failing tests but if it's easier for you that way then go ahead

mixmix:
no failing tests in the master branch IMO. will hold till your other thing is done.

index.js
```diff
@@ -32,14 +32,21 @@ const buildGroupId = require('./lib/build-group-id')
...
   manifest: {
     create: 'async',
-    publish: 'async',
+    get: 'async',
     list: 'source',
     addMembers: 'async',
+    excludeMembers: 'async',
+    publish: 'async',
+    start: 'async',
+    listMemebers: 'source',
```
staltz, line 47: @mixmix Typo here, "memebers"

index.js
```diff
@@ -32,14 +32,21 @@ const buildGroupId = require('./lib/build-group-id')
...
   name: 'tribes2',
   manifest: {
     create: 'async',
-    publish: 'async',
+    get: 'async',
     list: 'source',
     addMembers: 'async',
+    excludeMembers: 'async',
+    publish: 'async',
+    start: 'async',
```
staltz, line 46: `start` is now twice in this manifest

PR #84, opened 2023-04-05
lib/epoch - getMembers

mixmix:
A function which loads the membership of an epoch (along with those who have been marked for exclusion.

### DIFF
--- a/lib/epochs.js
+++ b/lib/epochs.js
@@ -3,7 +3,7 @@
 // SPDX-License-Identifier: LGPL-3.0-only
 
 const { promisify: p } = require('util')
-const { fromMessageSigil, fromFeedSigil } = require('ssb-uri2')
+const { fromMessageSigil } = require('ssb-uri2')
 const pull = require('pull-stream')
 const Reduce = require('@tangle/reduce')
 const OverwriteFields = require('@tangle/overwrite-fields')
@@ -13,7 +13,9 @@ const {
   validator: {
     group: {
       initRoot: isInitRoot,
-      initEpoch: isInitEpoch
+      initEpoch: isInitEpoch,
+      addMember: isAddMember,
+      exclude: isExclude,
     }
   }
 } = require('private-group-spec')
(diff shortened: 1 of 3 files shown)


mixmix:
@Powersource this and the #83 PR are both failing because exclusion isn't quite working yet. If this is a surprise, we should def talk!!

lib/tangles/get-tangle-updates.js
```diff
@@ -9,7 +9,16 @@ const B_CONTENT = Buffer.from('content')
...
 
-module.exports = function getUpdates(ssb, tangle, root, cb) {
+function getTangleUpdates(ssb, tangle, root, cb) {
+  pull(
+    tangleUpdateStream(ssb, tangle, root),
+    pull.collect(cb)
+  )
+}
+getTangleUpdates.stream = tangleUpdateStream
+module.exports = getTangleUpdates
```
mixmix, line 19: I just tweaked this file so you can get _EITHER_ a callback _OR_ a stream. :wink:

lib/epochs.js
```diff
@@ -108,6 +117,8 @@ function reduceEpochs (ssb, groupId, cb) {
             })
           }),
           pull.collect((err, nodes) => {
+            if (err) return cb(err)
```
Powersource, line 120: `clarify`ing errs would be nice

lib/tangles/get-tangle-updates.js
Powersource: aah that's what the .stream is for i was confused

lib/epochs.js
```diff
@@ -119,12 +130,32 @@ function reduceEpochs (ssb, groupId, cb) {
...
-function toFeedURI (id) {
-  if (id.startsWith('@')) return fromFeedSigil(id)
-  else return id
+
+function getEpochMembers (ssb, epochId, cb) {
+  const members = new Set()
+  const toExclude = new Set()
+
+  pull(
+    getTangleUpdates.stream(ssb, 'members', epochId),
```
Powersource, line 142: aah epochId is the root msg id of the epoch? i guessed that it was the epoch secret. maybe call it epochRoot?

Powersource (changes requested):
Looks nice just some small stuff

lib/epochs.js
mixmix: epochRoot is a good idea. In practice id === root for tangles, but this id is waaaaay overloaded as a term

Issue #79, opened 2023-03-24
Can't re-add people to a new epoch if they haven't posted to the old epoch, since triangulation fails


Powersource:
Found here https://github.com/ssbc/ssb-tribes2/pull/78#discussion_r1147598540

Powersource:
2. hopefully doesn't even bite us in prod, since people should only replicate additions feeds once we replicate metafeeds, and people will probably only do that once they themselves have groups/metafeeds and therefore called start() and have their own additions feed

Powersource:
#85 seems to have fixed all the issues in #78 

@mixmix you might be unblocked already. if not then you might have to wait for #78

PR #85, opened 2023-04-05
Publish re-ad messages on the additions feed

Powersource:
for https://github.com/ssbc/ssb-tribes2/issues/79 Hopefully fixes it, but we need to verify in #78 

tests were passing before the first commit

### DIFF
--- a/index.js
+++ b/index.js
@@ -177,91 +177,70 @@ module.exports = {
         // prettier-ignore
         if (err) return cb(clarify(err, "Couldn't get own root when excluding members"))
 
-        get(groupId, (err, { writeKey: oldWriteKey } = {}) => {
+        const excludeContent = {
+          type: 'group/exclude',
+          excludes: feedIds,
+          recps: [groupId],
+        }
+        const excludeOpts = {
+          tangles: ['members'],
+          isValid: isExclude,
+        }
+        publish(excludeContent, excludeOpts, (err) => {
           // prettier-ignore
-          if (err) return cb(clarify(err, "Couldn't get old key when excluding members"))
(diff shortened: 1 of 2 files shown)


Powersource:
@arj03 no it's re-add, no read :P

index.js
```diff
@@ -177,91 +177,70 @@ module.exports = {
...
-        get(groupId, (err, { writeKey: oldWriteKey } = {}) => {
+        const excludeContent = {
+          type: 'group/exclude',
+          excludes: feedIds,
+          recps: [groupId],
+        }
+        const excludeOpts = {
+          tangles: ['members'],
+          isValid: isExclude,
+        }
```
mixmix, line 188: I think we need to think about where we want the exclude published... (which feed).
I have a feeling it should be in the same feed as the additions... but I can't quite justify.
I think it's that we want to be sure we have all the messages for each epoch membership stuff and sprinkling the around might lead to funny edge cases?

what do you think?

test/exclude-members.test.js
```diff
@@ -160,6 +145,32 @@ test('add and remove a person, post on the new feed', async (t) => {
 
   t.equal(post.text, 'post', 'found post on second feed')
 
+  const aliceAdditions = await p(alice.metafeeds.findOrCreate)({
+    purpose: 'group/additions',
+  })
+  const msgsFromAdditions = await alice.db.query(
+    where(author(aliceAdditions.id)),
+    toPromise()
+  )
```
mixmix, line 154: :pineapple: it's cool seeing different ways of querying around the space + testing. I would have got a different way but this is tidy (looking within a particular sub-feed).
db.query is so sweet

Issue #71, opened 2023-03-16
Test being removed from a group


Powersource:
Maybe could simply check that we can't read any of the re-add messages

Should we remove it from the list of groups or something? Maybe not remove it but mark it in some way?

I'm guessing this is what a removed member will see, at least for the group profile 
![image](https://user-images.githubusercontent.com/1427063/225601407-10e325d3-d097-4ca4-9541-6baca97f55e2.png)
https://www.figma.com/file/y96kXBNrjb7pQHiM3mqz6x/NGI-group-profile?t=4DzXriMlLjp0lomb-0

But that's maybe only the door. What about the group contents? Ask nonlinear.

Powersource:
we concluded stuff in chat

> We did it so Alice just return to a non member. She sees just what a non member would.

> nonlinear: do ex members experience groups as non members? As in, they see the door, etc? Or they experience it differently? (Simpler is yes, same experience)
> andré: I think they should just go back to being treated as non-members, so they just see the door. Additionally some kind of UI that tells they were excluded from the group.
> nonlinear: they receive a notification

Powersource:
update: there should be some note that you've been removed, so you don't feel gaslit

PR #88, opened 2023-04-07
Add test for being removed from a group

Powersource:
Fixes https://github.com/ssbc/ssb-tribes2/issues/71

* [x] what happens if we restart our client, will we rejoin groups we've left? new issue probably. made one: https://github.com/ssbc/ssb-tribes2/issues/90
* [x] test that listinvites doesn't list the old invite
* [x] test listMembers. what should it return?
* [x] make sure stuff works once #78 is merged.

### DIFF
--- a/index.js
+++ b/index.js
@@ -450,13 +450,14 @@ module.exports = {
             live({ old: true }),
             toPullStream()
           ),
+          pull.filter(isExclude),
           pull.filter((msg) =>
             // it's an exclusion of us
-            msg.value?.content?.excludes?.includes(myRoot.id)
+            msg.value.content.excludes.includes(myRoot.id)
           ),
           pull.drain(
             (msg) => {
-              const groupId = msg.value?.content?.recps?.[0]
+              const groupId = msg.value.content.recps[0]
               ssb.box2.excludeGroupInfo(groupId, null)
             },
             (err) => {
--- a/index.js
+++ b/index.js
@@ -442,8 +442,9 @@ module.exports = {
 
       ssb.metafeeds.findOrCreate((err, myRoot) => {
         // prettier-ignore
-        if (err) return cb(clarify(err, 'todo'))
+        if (err) return cb(clarify(err, 'Error getting own root in start()'))
 
+        // check if we've been excluded
         pull(
           ssb.db.query(
             where(and(isDecrypted('box2'), type('group/exclude'))),
@@ -462,17 +463,12 @@ module.exports = {
             },
             (err) => {
               // prettier-ignore
-              if (err) return cb(clarify(err, 'todo'))
+              if (err) return cb(clarify(err, 'Error on looking for exclude messages excluding us'))
             }
           )
         )
-      })
 
-      // look for new epochs that we're added to
-      ssb.metafeeds.findOrCreate((err, myRoot) => {
-        // prettier-ignore
-        if (err) return cb(clarify(err, 'Error getting own root in start()'))
-
+        // look for new epochs that we're added to
         pull(
           ssb.db.query(
             // TODO: does this output new stuff if we accept an invite to an old epoch and then find additions to newer epochs?
--- a/index.js
+++ b/index.js
@@ -5,6 +5,7 @@
 const { promisify } = require('util')
 const pull = require('pull-stream')
 const paraMap = require('pull-paramap')
+const pullMany = require('pull-many')
 const lodashGet = require('lodash.get')
 const clarify = require('clarify-error')
 const {
@@ -117,7 +118,13 @@ module.exports = {
     }
 
     function list(opts = {}) {
-      return pull(ssb.box2.listGroupIds({ live: !!opts.live }), paraMap(get, 4))
+      return pull(
+        ssb.box2.listGroupIds({
+          live: !!opts.live,
+          excluded: !!opts.excluded,
+        }),
+        paraMap(get, 4)
+      )
     }
 
     function addMembers(groupId, feedIds, opts = {}, cb) {
@@ -346,7 +353,11 @@ module.exports = {
             if (err) return cb(clarify(err, 'Failed to get root metafeed when listing invites'))
 
             pull(
-              ssb.box2.listGroupIds(),
+              pullMany([
+                ssb.box2.listGroupIds(),
+                ssb.box2.listGroupIds({ excluded: true }),
+              ]),
+              pull.flatten(),
               pull.collect((err, groupIds) => {
                 // prettier-ignore
                 if (err) return cb(clarify(err, 'Failed to list group IDs when listing invites'))
--- a/index.js
+++ b/index.js
@@ -297,16 +297,36 @@ module.exports = {
 
     function listMembers(groupId, opts = {}) {
       return pull(
-        ssb.db.query(
-          where(and(isDecrypted('box2'), type('group/add-member'))),
-          opts.live ? live({ old: true }) : null,
-          toPullStream()
-        ),
-        pull.map((msg) => lodashGet(msg, 'value.content.recps', [])),
-        pull.filter((recps) => recps.length > 1 && recps[0] === groupId),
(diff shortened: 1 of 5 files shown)


test/exclude-members.test.js
```diff
@@ -174,3 +174,74 @@ test('add and remove a person, post on the new feed', async (t) => {
...
+      type: 'test',
+      text: 'bob second post',
+      recps: [groupId],
+    })
+    .then(() =>
+      t.fail("Bob posted again in the group even if he's removed from it")
+    )
+    .catch(() =>
+      t.pass("Bob can't post in the group anymore since he's removed from it")
+    )
```
Powersource, line 244: Make an issue to make a test where we don't replicate here, so bob still can post, but once they try to replicate, alice ignores bob's new post because we're limiting replication. or hmm that's for replication scheduler i guess.

Powersource:
adapting keyring https://gitlab.com/ahau/lib/ssb-keyring/-/merge_requests/23

Powersource:
https://github.com/ssbc/ssb-box2/pull/28

package.json
```diff
@@ -43,7 +43,7 @@
     "pull-paramap": "^1.2.2",
     "pull-stream": "^3.7.0",
     "ssb-bfe": "^3.7.0",
-    "ssb-box2": "^6.0.0",
+    "ssb-box2": "github:ssbc/ssb-box2#remove-group-info",
```
Powersource, line 46: https://github.com/ssbc/ssb-box2/pull/28

index.js
```diff
@@ -407,6 +407,34 @@ module.exports = {
...
+      ssb.metafeeds.findOrCreate((err, myRoot) => {
+        // prettier-ignore
+        if (err) return cb(clarify(err, 'todo'))
+
+        pull(
+          ssb.db.query(
+            where(and(isDecrypted('box2'), type('group/exclude'))),
+            live({ old: true }),
+            toPullStream()
+          ),
```
Powersource, line 420: I'm a bit worried how this will interact with listInvites (hmm that function should check if `removed: true`) and epoch discovery #78 (i guess that should also check `removed`)

I'll add todos for those.

Will this interact well if we get added back into a group though?

Powersource: andre and i chatted a bit

> I haven't looked at code, but couldn't this be discovered with epoch data?
> If you're invited to epoch N, it doesn't matter if you were excluded or not on epoch X<N

yeah this is maybe another case where we need preferredEpoch
only care about an exclude if it's in the preferredEpoch

> Yeah
> For implementing the fork-merging rules we might anyway need to do more epoch kung fu, so we need the basic epoch-related queries for that
> Not sure if I make sense

or maybe not necessarily preferredEpoch, since you can
1. get added to and removed from epoch 0
2. get an invite to epoch 1
3. epoch 1 rotates to epoch 2, you get re-added to it
4. you accept the invite to epoch 1, even though epoch 2 is preferred

test/exclude-members.test.js
```diff
@@ -174,3 +174,80 @@ test('add and remove a person, post on the new feed', async (t) => {
...
```
Powersource, line 247: wut

```
ok 28 bob posts in the group while he's in it
removing for groupid ssb:identity/group/fq4HgHIJLXxZw78EQqziJ4QL5rsTpE2TvtiRL8ruyhI=
node:assert:400
    throw err;
    ^

AssertionError [ERR_ASSERTION]: The expression evaluated to a falsy value:

  assert(node.key)

    at isRoot (/home/me/prj/ssb/ssb-tribes2/node_modules/.pnpm/@tangle+graph@3.2.0/node_modules/@tangle/graph/lib/is-root.js:4:3)
    at /home/me/prj/ssb/ssb-tribes2/node_modules/.pnpm/@tangle+graph@3.2.0/node_modules/@tangle/graph/index.js:18:23
    at Array.filter (<anonymous>)
    at Graph.setupGraph (/home/me/prj/ssb/ssb-tribes2/node_modules/.pnpm/@tangle+graph@3.2.0/node_modules/@tangle/graph/index.js:18:8)
    at new Graph (/home/me/prj/ssb/ssb-tribes2/node_modules/.pnpm/@tangle+graph@3.2.0/node_modules/@tangle/graph/index.js:8:10)
    at new Reduce (/home/me/prj/ssb/ssb-tribes2/node_modules/.pnpm/@tangle+reduce@5.0.4/node_modules/@tangle/reduce/index.js:13:18)
    at /home/me/prj/ssb/ssb-tribes2/lib/tangles/get-tangle-data.js:94:22
    at /home/me/prj/ssb/ssb-tribes2/node_modules/.pnpm/pull-stream@3.7.0/node_modules/pull-stream/sinks/reduce.js:10:5
    at /home/me/prj/ssb/ssb-tribes2/node_modules/.pnpm/pull-stream@3.7.0/node_modules/pull-stream/sinks/drain.js:26:24
    at /home/me/prj/ssb/ssb-tribes2/node_modules/.pnpm/pull-stream@3.7.0/node_modules/pull-stream/throughs/filter.js:17:11 {
  generatedMessage: true,
  code: 'ERR_ASSERTION',
  actual: undefined,
  expected: true,
  operator: '=='
}
```

index.js
Powersource: > By the time you accept invite to epoch 1, then epoch 1 is preferred at *that* time
> As soon as you discover epoch 2, then that switches
> Preferred epoch is a subjective local-peer concept, it's not global for all peers
> But by design we want them to converge to the same

test/exclude-members.test.js
```diff
@@ -174,3 +174,84 @@ test('add and remove a person, post on the new feed', async (t) => {
...
+
+  // TODO: check bob's group list, it should still have an entry for the group but it should be marked that he's removed from it or something. or an opt for the function?
```
Powersource, line 250: todo

Powersource: can probably just straight up return what we get from box2. opts can be added later if people want the list pre-filtered

test/exclude-members.test.js
```diff
@@ -174,3 +174,84 @@ test('add and remove a person, post on the new feed', async (t) => {
...
+
+  // TODO: try to check the messages encrypted to the new key/epoch, and fail
+  // checking the re-addition messages on alice's additions feed should do
```
Powersource, line 253: if we can decrypt a message then we should have done that by now. get the re-add msg id somehow and check that when bob gets it it's still encrypted (still a string)

test/exclude-members.test.js
Powersource: thought i fixed this one but apparently not

Powersource: lmao forgot the remove rename

index.js
Powersource: made issue for this https://github.com/ssbc/ssb-tribes2/issues/89

test/exclude-members.test.js
Powersource: yeah that's a replication concern, shouldn't be directly tracked here. made a task for me in airtable

Powersource:
I want to wait for #78 first before merging but this should be mostly reviewable now.

test/exclude-members.test.js
```diff
@@ -174,3 +175,114 @@ test('add and remove a person, post on the new feed', async (t) => {
...
+    )
+    .catch(() =>
+      t.pass("Bob can't post in the group anymore since he's excluded from it")
+    )
+
+  const bobGroups = await pull(bob.tribes2.list(), pull.collectAsPromise())
+  t.deepEquals(
+    bobGroups,
+    [{ id: groupId, excluded: true }],
+    "bob is excluded from the only group he's been in. sad."
```
mixmix, line 261: :laughing:

mixmix (approved):
Beautiful

# :bouquet:

Powersource:
@SocketSecurity ignore-all

PR #91, opened 2023-04-14
Epoch check missing members

mixmix:
adds the ability to spot members missing.

**Minimal test at the moment:** hack so that a person is missed from being added to the new epoch. This detects that.

Algorithm: 
- for each epoch, collect all the people who have been added/ exluded up to that epoch
- check the group members are that group (added sofar - excluded sofar)
- if not, then note those members missing

COULD be good to test the forked case...

### DIFF
--- a/lib/epochs.js
+++ b/lib/epochs.js
@@ -31,6 +31,13 @@ function toPattern (regexp) {
   return regexp.toString().replace(/^\//, '').replace(/\/$/, '')
 }
 
+// This strategy describes how to "reduce" the tangle of epochs and their data.
+// Here OverwriteFieleds lets any link in the tangle contain Objects with form
+// { [key]: value } so long as
+//   - key matches msgPattern,
+//   - value is { author, epochKey }
+//
+// Since each key is unique here this behaves like Object.assign (with checks)
 const strategy = OverwriteFields({
   keyPattern: msgPattern,
   valueSchema: {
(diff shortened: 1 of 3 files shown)


Powersource:
woa intense. i'll try again to read this next week

lib/epochs.js
```diff
@@ -20,109 +20,211 @@ const {
...
+
+              if (historyNode.key === node.key) return
+              // INFO node members toExclude is talking about what should happen
+              // *after* it in the graph, so if we're enquiring about a
+              // particular epoch, we don't include that epoch's toExclude
+              toExclude.forEach(feedId => excludedSoFar.add(feedId))
+            })
+
+          // Check if those who should be present are present
+          const shouldBePresent = difference(addedSoFar, excludedSoFar)
```
Powersource, line 139: Doesn't this ignore people who have been removed and added back?

lib/epochs.js
```diff
@@ -20,109 +20,211 @@ const {
...
+          const currentMembers = new Set(node.data[node.key].members.added)
+          if (isSameSet(shouldBePresent, currentMembers)) return acc
+
+          // If find some missing, record this epoch and missing members
+          acc.push({
+            epoch: node.key, // alias: epochRootId
+            epochKey: Buffer.from(node.data[node.key].epochKey, 'base64'),
```
Powersource, line 146: epochRootId?

Powersource: no wait this is the secret. but a rename to something else would be helpful

lib/epochs.js
```diff
@@ -20,109 +20,211 @@ const {
...
+      getMembers(ssb, epochRootId, cb)
+    },
+
+    getMissingMembers (groupId, cb) {
+      if (cb === undefined) return p(this.getMissingMembers)(groupId)
+
+      reduceEpochs(ssb, groupId, allGetters, (err, reduce) => {
+        if (err) return cb(clarify(err, 'Failed to resolve epoch @tangle/reduce'))
+
+        if (reduce.graph.connectedNodes.length === 1) return cb(null, [])
```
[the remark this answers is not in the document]
mixmix: I've documented it now for you.
I hadn't done this before now because it is not save for people to manipulate it ... see the docs

lib/epochs.js
mixmix: yeah we're not covering that case for the moment. LATER! (/never)

lib/epochs.js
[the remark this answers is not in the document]
Powersource: still not sure what connectedNodes is (doesn't seem documented in graph) but this is clearer, cool

Powersource (approved):

PR #92, opened 2023-04-14
lib/epochs tieBreak

mixmix:
Feel free to merge

### DIFF
--- a/lib/epochs.js
+++ b/lib/epochs.js
@@ -151,6 +151,20 @@ module.exports = function Epochs (ssb) {
 
         cb(null, result)
       })
+    },
+
+    tieBreak (epochs) {
+      if (!epochs || !Array.isArray(epochs))
+        throw Error('tieBreak requires an Array of epochs')
+
+      const keys = epochs
+        .map(epoch => epoch.epochKey.toString('hex'))
+        .sort()
+      console.log(keys)
+
+      const winningKey = Buffer.from(keys[0], 'hex')
+      
+      return epochs.find(epoch => epoch.epochKey.equals(winningKey))
     }
   }
 }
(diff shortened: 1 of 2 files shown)


lib/epochs.js
```diff
@@ -151,6 +151,19 @@ module.exports = function Epochs (ssb) {
...
       })
+    },
+
+    tieBreak (epochs) {
+      if (!epochs || !Array.isArray(epochs))
+        throw Error('tieBreak requires an Array of epochs')
+
+      const keys = epochs
+        .map(epoch => epoch.epochKey.toString('base64'))
+        .sort()
```
Powersource, line 162: https://github.com/ssbc/ssb-group-exclusion-spec#43-tie-breaking-rule


> The tie-breaking rule is: among the forked epochs, sort their epoch keys by lexicographic order (in hexadecimal encoding) and the epoch with the smallest epoch key is considered the winner.


Needs to be hex encoding not base64!

Powersource (changes requested):

lib/epochs.js
```diff
@@ -151,6 +151,19 @@ module.exports = function Epochs (ssb) {
...
         cb(null, result)
       })
+    },
+
+    tieBreak (epochs) {
+      if (!epochs || !Array.isArray(epochs))
+        throw Error('tieBreak requires an Array of epochs')
+
+      const keys = epochs
+        .map(epoch => epoch.epochKey.toString('base64'))
```
mixmix, line 161:
```suggestion
        .map(epoch => epoch.epochKey.toString('hex'))
```

mixmix:
updated, please review again and merge @Powersource

Issue #86, opened 2023-04-07
Test getting added to an old epoch but still finding the newer epochs we've been re-added to


Powersource:
Followup from https://github.com/ssbc/ssb-tribes2/pull/78#discussion_r1159460854

Powersource:
like 

1. A invites B to epoch 0.
2. A excludes C and creates epoch 1.
3. B accepts invite to epoch 0.
4. B should automatically discover and use epoch 1.

Issue #87, opened 2023-04-07
Don't always automatically switch to an epoch when finding it


Powersource:
Making this logic more refined https://github.com/ssbc/ssb-tribes2/pull/78#discussion_r1159461684

Idk if this is a concrete todo yet or if this will maybe resolve itself when we fix other issues, as well as @mixmix working on the "preferred epoch" logic

Powersource:
another case https://github.com/ssbc/ssb-tribes2/pull/94/files#r1168847046

PR #94, opened 2023-04-17
Test being added to an old epoch

Powersource:
Fixes https://github.com/ssbc/ssb-tribes2/issues/86

* [x] remove the old todo comment


index.js
```diff
@@ -392,10 +394,32 @@ module.exports = {
...
                       root: lodashGet(msg, 'value.content.root'),
                     }
-                  })
-                )
+                  }),
+                  // aggregate multiple readKeys from invites to different epochs into one invite object
+                  pull.drain(
+                    (invite) => {
+                      // TODO: which writeKey should be picked??
+                      // or is that maybe not that important to decide here, maybe for acceptInvite?
```
Powersource, line 402: @mixmix ran into another case where we need to pick (just fyi)

Powersource: related https://github.com/ssbc/ssb-tribes2/issues/87

index.js
```diff
@@ -415,6 +439,7 @@ module.exports = {
         pull.drain(
           (groupInfo) => {
             foundInvite = true
+            // TODO: loop over all the readKeys we've found and add them all
```
Powersource, line 442: todo

index.js
```diff
@@ -392,10 +394,32 @@ module.exports = {
                       readKeys: [{ key, scheme }],
                       root: lodashGet(msg, 'value.content.root'),
                     }
-                  })
-                )
+                  }),
+                  // aggregate multiple readKeys from invites to different epochs into one invite object
+                  pull.drain(
```
Powersource, line 399: this won't be very streamy now, but maybe that's ok for now since we're just getting it to work at first?

index.js
Powersource: done. note that the "bob managed to decrypt alice's second message" test was failing before i changed this, even with a decent timeout, which means epochs weren't rediscovered by start() (i think?)

index.js
Powersource: moved this comment down to acceptInvite now

Powersource: tracking this in #87 instead

index.js
```diff
@@ -482,7 +514,6 @@ module.exports = {
         // look for new epochs that we're added to
         pull(
           ssb.db.query(
-            // TODO: does this output new stuff if we accept an invite to an old epoch and then find additions to newer epochs?
```
Powersource, line 485: answer seems to be no, but it's fine since acceptInvite now consumes all epochs it can find we've been invited to

mixmix (changes requested):
The test is good. The multicb got under my skin so I did a little refactor I'd like to recommend instead:

Please see this PR @Powersource  https://github.com/ssbc/ssb-tribes2/pull/96

package.json
```diff
@@ -40,10 +40,11 @@
     "envelope-spec": "^1.1.1",
     "fast-deep-equal": "^3.1.3",
     "is-canonical-base64": "^1.1.1",
+    "jitdb": "^7.0.7",
```
Powersource, line 43: had to add this as tests were failing when i pulled in master because of missing jitdb/operators dep. but when installing this i got

```
% pnpm add jitdb
Packages: +1
+
Progress: resolved 521, reused 521, downloaded 0, added 1, done

dependencies:
+ jitdb 7.0.7

 WARN  Issues with peer dependencies found
.
└─┬ jitdb 7.0.7
  └── ✕ missing peer async-append-only-log@^4.3.2
Peer dependencies that should be installed:
  async-append-only-log@^4.3.2

Done in 5.9s
```

things seem to work though

@mixmix

Issue #77, opened 2023-03-23
Adapt listMembers to exclusion


Powersource:
I guess the excluded member will have to be removed from the list. 

Probably too early to worry about how to list old members. (maybe just search for exclude messages and subtract current members?)

Powersource:
specifically https://github.com/ssbc/ssb-tribes2/pull/84

PR #95, opened 2023-04-18
Adapt listMembers to exclusion and test it

Powersource:
Fixes https://github.com/ssbc/ssb-tribes2/issues/77


index.js
```diff
@@ -310,6 +310,7 @@ module.exports = {
     }
 
     function listMembers(groupId, opts = {}) {
+      // TODO: getPickedEpoch -> getMembers for that epoch
```
Powersource, line 313: the plan

index.js
```diff
@@ -313,31 +313,16 @@ module.exports = {
...
 
-            if (group.excluded) {
-              return cb(
-                new Error("We're excluded from this group, can't list members")
-              )
-            } else {
-              const source = pull(
-                ssb.db.query(
-                  where(and(isDecrypted('box2'), type('group/add-member'))),
-                  opts.live ? live({ old: true }) : null,
```
Powersource, line 328: Realized that we'll need to support `opts.live` (or?) which makes this pr trickier. But I think it should be possible without too much work, box2/keyring will have to have a live `get` when changes happen, e.g. new picked write key, and getMembers will have to support live as well, but that doesn't look that hard either.

test/exclude-members.test.js
```diff
@@ -413,6 +413,26 @@ test("If you're not the excluder nor the excludee then you should still be in th
...
+  const carolMembers = await pull(
+    carol.tribes2.listMembers(groupId),
+    pull.collectAsPromise()
+  )
+  t.deepEquals(
+    carolMembers.sort(),
+    [aliceRoot.id, carolRoot.id].sort(),
+    'carol gets the correct members list'
+  )
+
```
Powersource, line 435: This might make more sense to move into list-members.test.js, we can't put absolutely everything into this file, exclusion will touch too much

test/exclude-members.test.js
Powersource, line 1: Need to add `live` test, makes more sense in list-members.test.js

lib/tangles/get-tangle-data.js
```diff
@@ -27,6 +27,7 @@ function getTangleRoot(server, tangle, groupId, cb) {
       return cb(new Error(`get-tangle: unknown groupId ${groupId}`))
     }
 
+    // TODO: use getPickedEpoch?
```
Powersource, line 30: make a separate issue for this

package.json
```diff
@@ -49,7 +49,7 @@
     "pull-stream": "^3.7.0",
     "set.prototype.difference": "^1.0.2",
     "ssb-bfe": "^3.7.0",
-    "ssb-box2": "^6.1.0",
+    "ssb-box2": "ssbc/ssb-box2#Powersource-patch-1",
```
Powersource, line 52: https://github.com/ssbc/ssb-box2/pull/31

lib/tangles/get-tangle-data.js
Powersource: we actually check against the current writeKey here, i missed that when skimming, so this function should be fine actually

index.js
```diff
@@ -319,35 +319,19 @@ module.exports = {
...
-        deferedSource.resolve(source)
-      })
-
-      return deferedSource
+      return pull(
+        getPickedEpoch(groupId, { live: !!opts?.live }),
+        pull.map((pickedEpoch) =>
+          getMemberUpdates(pickedEpoch.id, { live: !!opts?.live })
+        ),
+        pull.flatten(),
```
mixmix, line 327: I honestly don't know what flaten does if it's operating on streams... especially if the streams are "live"... if there are e.g. 2 epochs live, does the second epoch ever get a word in ?

README.md
```diff
@@ -162,7 +162,7 @@ Publishes any kind of message encrypted to the group. The function wraps `ssb.db
 
 ### `ssb.tribes2.listMembers(groupId, { live }) => source`
 
-Returns a pull stream source listing every known member of the group with id `groupId`. Note: lists members whether or not they've accepted the invite. If `live` is true, then it keeps the stream open and also outputs new members.
+Returns a pull stream source listing the root feed id of every member of the group with id `groupId`. Note: lists members whether or not they've accepted the invite. If `live` is true, then it keeps the stream open and also outputs new members. If `live` is true and a member gets excluded, it outputs an object on the format `{ excluded: excludedPeerRootFeedId }`.
```
mixmix, line 165: I personally really dislike streams which have varied output like this.
in the live case you're forcing the person to handle a new type.

I think you should just leave the output in `{ added: [], excluded[] }` form, then the fields are always there, and instead of going `isString or isObject, then do this or that` the person can straight up just up `update.added` and trust it's always there.

mixmix: Also this might be pet peeve and I will not die if this is merged, I will just be annoyed whenever I see type checking

mixmix: why is it annoying? it feels like a footgun. You decide to change to "live" and your stream pipeline that was working now explodes

lib/epochs.js
```diff
@@ -78,27 +78,72 @@ module.exports = function Epochs(ssb) {
     },
   }
 
+  function getEpochs(groupId, cb) {
+    if (cb === undefined) return p(this.getEpochs)(groupId)
```
mixmix, line 82:
```suggestion
    if (cb === undefined) return p(getEpochs)(groupId)
```

lib/epochs.js
```diff
@@ -78,27 +78,72 @@ module.exports = function Epochs(ssb) {
...
-            ...node.data[node.key],
-          }
-          info.secret = Buffer.from(info.secret, 'base64')
-          return info
+    getEpochs,
+
+    /**
+     * Get the epoch matching the currently picked write key for the group.
+     */
+    getPickedEpoch(groupId, opts) {
```
mixmix, line 108: This function is already one I wrote in https://github.com/ssbc/ssb-tribes2/pull/99 ... :cry: 
I'm gonna get that merged and you should use that because I'm writing legit tests around it based on the specs

mixmix: :fire: TO CHANGE

mixmix: also I called it `getPreferredEpoch` to match the name in the spec

test/list-members.test.js
```diff
@@ -138,3 +138,136 @@ test('live list members', async (t) => {
...
+    )
+    .catch(() =>
+      t.pass(
+        "Bob gets an error when trying to list members of the group he's excluded from"
+      )
+    )
+
+  await p(alice.close)(true)
+  await p(bob.close)(true)
+  await p(carol.close)(true)
```
mixmix, line 272:
```suggestion
  await Promise.all([
    p(alice.close)(true),
    p(bob.close)(true),
    p(carol.close)(true),
  ])
```

mixmix (changes requested):
Good tests.

Changes requested:
- [x] change the API of `listMembers` to always emit the same shaped objects (medium strength request)
- [ ] change to use the API I am writing for `getPreferredEpoch` (strong request)

README.md
Powersource: ah i guess that's an option, to have it on object form. i think i wanted more backwards compat. but

1. do you think it should have this object form only on `live` or also when not live? because not live i think excluded would always be empty (if it gets filled you get a new epoch) although that brings me to my second question...
2. should the `added` list be a list of the members added just now or every time it's emitted it should be a complete list of all members? if it's always a complete list then we could just skip the `excluded` list and just return `Stream<AllCurrentMembers[]>`. Since the list is very unlikely to be >200 member ids long it wouldn't need to be a stream of a stream right? Although `Stream<AllCurrentMembers[]>` would just be a 1 item long stream when not live, feels weird?

index.js
Powersource: noo it fails. nice catch

lib/epochs.js
```diff
@@ -78,27 +78,72 @@ module.exports = function Epochs(ssb) {
...
+
+          lastKey = groupInfo.writeKey.key
+          return true
+        }),
+        // find details about the newly picked epoch
+        pull.asyncMap((groupInfo, cb) => {
+          getEpochs(groupId, (err, epochs) => {
+            // prettier-ignore
+            if (err) return cb(clarify(err, "Error getting info about epochs when looking for picked epoch"))
+
```
Powersource, line 133: was getting errors so added a log on this line
```
            console.log('got epochs but no err', epochs, 'groupInfo', groupInfo)
```

and got these logs. it's flaky though so happens on like 50% of runs

```
...
got epochs but no err [
  {
    id: 'ssb:message/classic/HQJyiWkVCVosHwKQPmfYkcShF0jWZOBnozNQQ6fPTc0=',
    previous: null,
    author: 'ssb:feed/bendybutt-v1/R7oVP6uSBqSmyC-iQxX67S9Zl2498yxJk4NarqMaBtI=',
    secret: <Buffer 39 59 6c 7d fc 27 3d d7 49 42 ec ec 13 77 5b ae 5d b4 93 17 9c 48 c1 05 8a 40 01 79 2c a2 dc 8c>
  }
] groupInfo {
  readKeys: [
    {
      key: <Buffer 39 59 6c 7d fc 27 3d d7 49 42 ec ec 13 77 5b ae 5d b4 93 17 9c 48 c1 05 8a 40 01 79 2c a2 dc 8c>,
      scheme: 'envelope-large-symmetric-group'
    },
    {
      key: <Buffer 6b bd f6 bb 5f f3 61 58 dd 46 3d 65 63 e1 4d e0 3e 68 76 77 6a 4d 95 a6 35 7b 01 f4 4e 83 c7 10>,
      scheme: 'envelope-large-symmetric-group'
    }
  ],
  writeKey: {
    key: <Buffer 6b bd f6 bb 5f f3 61 58 dd 46 3d 65 63 e1 4d e0 3e 68 76 77 6a 4d 95 a6 35 7b 01 f4 4e 83 c7 10>,
    scheme: 'envelope-large-symmetric-group'
  },
  root: 'ssb:message/classic/HQJyiWkVCVosHwKQPmfYkcShF0jWZOBnozNQQ6fPTc0='
}
pickedEPoch undefined
not ok 5 TypeError: Cannot read properties of undefined (reading 'id')
  ---
    operator: error
    at: <anonymous> (/home/me/prj/ssb/ssb-tribes2/test/list-members.test.js:204:18)
    stack: |-
      TypeError: Cannot read properties of undefined (reading 'id')
          at /home/me/prj/ssb/ssb-tribes2/index.js:326:47
          at /home/me/prj/ssb/ssb-tribes2/node_modules/.pnpm/pull-stream@3.7.0/node_modules/pull-stream/throughs/map.js:13:23
          at /home/me/prj/ssb/ssb-tribes2/node_modules/.pnpm/pull-stream@3.7.0/node_modules/pull-stream/throughs/async-map.js:39:20
          at /home/me/prj/ssb/ssb-tribes2/lib/epochs.js:137:13
          at /home/me/prj/ssb/ssb-tribes2/lib/epochs.js:98:7
          at /home/me/prj/ssb/ssb-tribes2/lib/epochs.js:333:13
          at /home/me/prj/ssb/ssb-tribes2/node_modules/.pnpm/pull-stream@3.7.0/node_modules/pull-stream/sinks/reduce.js:10:5
          at /home/me/prj/ssb/ssb-tribes2/node_modules/.pnpm/pull-stream@3.7.0/node_modules/pull-stream/sinks/drain.js:26:24
          at /home/me/prj/ssb/ssb-tribes2/node_modules/.pnpm/pull-stream@3.7.0/node_modules/pull-stream/throughs/async-map.js:28:19
          at /home/me/prj/ssb/ssb-tribes2/node_modules/.pnpm/pull-stream@3.7.0/node_modules/pull-stream/sources/values.js:19:7
  ...
```

it's weird that getEpochs returns fewer epochs/secrets than getGroupInfoUpdates. a bug in getEpochs? @mixmix

index.js
Powersource: fixed this with a cool pull.take. things are a bit flaky though, see other thread

Powersource:
"request review" meaning more like i have questions pls reply

index.js
```diff
@@ -319,48 +325,59 @@ module.exports = {
...
       getMyGroups((err, myGroups) => {
         // prettier-ignore
-        if (err) return deferedSource.abort(clarify(err, 'Failed to list group IDs when listing invites'))
+        if (err) return deferredSource.abort(clarify(err, 'Failed to list group IDs when listing invites'))
 
         const source = pull(
           // get all the groupIds we've heard of from invites
           ssb.db.query(
-            where(and(isDecrypted('box2'), type('group/add-member'))),
+            where(or(isDecrypted('box2'), type('group/add-member'))),
```
mixmix, line 380:
```suggestion
            where(and(isDecrypted('box2'), type('group/add-member'))),
```

mixmix:
@Powersource I merged my refactor in (after putting README in, and the shape of the data coming out of `listMembers`)
I also pulled all the listening functions that were running in `start()` out into `./listeners.js`  (see comment below)

index.js
```diff
@@ -481,97 +499,8 @@ module.exports = {
...
-              })
-            },
-            (err) => {
-              // prettier-ignore
-              if (err) return cb(clarify(err, "Error finding new epochs we've been added to"))
-            }
-          )
-        )
+        cb(null)
+        startListeners(ssb, getPreferredEpoch, console.error)
```
mixmix, line 503: BOOM, look at this cute lil guy. Tidies up a whole mess.
Also takes an error-handler at the end, so we can easily decide what we wanna do with those errors later and change this in one place

mixmix:
Finally... I found a limitation which is causing that test to fail.
Solution in this PR : https://github.com/ssbc/ssb-tribes2/pull/112

index.js
Powersource: woo nice

listeners.js
```diff
@@ -0,0 +1,179 @@
...
+
+            getPreferredEpoch(groupId, (err, epochInfo) => {
+              // prettier-ignore
+              if (err && !isClosed) return onError(clarify(err, 'fatal error getting preferred epoch'))
+              if (groupInfo.writeKey.key.equals(epochInfo.secret)) return
+
+              ssb.box2.addGroupInfo(
+                groupId,
+                { key: epochInfo.secret },
+                (err) => {
```
Powersource, line 150: I think this is unlikely to be needed here, we would need to have already added the epoch secret to figure out that the epoch is preferred i think

Powersource: but doesn't technically hurt and we'll be poking around here more later

test/exclude-members.test.js
```diff
@@ -275,18 +275,6 @@ test('Verify that you actually get excluded from a group', async (t) => {
...
-      t.fail(
-        "Bob didn't get an error when trying to list members of the group he's excluded from"
-      )
-    )
-    .catch(() =>
-      t.pass(
-        "Bob gets an error when trying to list members of the group he's excluded from"
-      )
-    )
-
```
Powersource, line 289: why remove this?

Powersource: ah  itäs moved

Powersource:
who's even reviewing this anymore lol? i'll just merge now lmao

PR #98, opened 2023-04-19
Promise.all setup/teardown in tests

mixmix:
Before: 342.403 s
After: 312.41 s

(on my desktop)

seems worth it


mixmix:
this is just me messing around in the evening (off the clock)

Powersource (approved):

PR #99, opened 2023-04-20
lib/epochs preferredEpoch case 4.4 (2 epochs)

mixmix:
I'm gonna do this in parts, because it's gonna feel overwhelming otherwise.

This covers the **case 4.4 - forked group with epochs with identical membership**
https://github.com/ssbc/ssb-group-exclusion-spec/#44-resolving-forked-epochs-with-same-membership

NOTE this does not cover the case of >2 fork tips

### DIFF
--- a/test/lib/epochs.test.js
+++ b/test/lib/epochs.test.js
@@ -36,6 +36,13 @@ function Run (t) {
   }
 }
 
+function getRootIds (peers) {
+  return Promise.all(
+    peers.map(peer => p(peer.metafeeds.findOrCreate)())
+  )
+    .then(feeds => feeds.map(feed => feed.id))
+}
+
 test('lib/epochs (getEpochs, getMembers)', async t => {
   const run = Run(t)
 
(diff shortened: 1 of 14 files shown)


test/lib/epochs.test.js
```diff
@@ -256,3 +260,86 @@ test('lib/epochs (tieBreak)', async t => {
...
+    Promise.all([
+      bob.tribes2.excludeMembers(group.id, [aliceId], {}),
+      oscar.tribes2.excludeMembers(group.id, [aliceId], {})
+    ])
+  )
+
+  await run('(sync exclusion)', replicate(bob, oscar))
+
+  const epochs = await Epochs(oscar).getEpochs(group.id)
+    .then(epochs => epochs.filter(epoch => epoch.author != aliceId))
```
Powersource, line 313: shouldn't this work even without filtering out the initial epoch that alice created?

Powersource: ah it's because it can only handle 2. i guess fine for now

test/lib/epochs.test.js
```diff
@@ -256,3 +260,86 @@ test('lib/epochs (tieBreak)', async t => {
...
+test.skip('lib/epochs (getPreferredEpoch - 4.5. subset membership)', async t => {
+  // the choice is the epoch which is a subset of all the others
+
+  t.end()
+})
+
+test.skip('lib/epochs (getPreferredEpoch - 4.6. overlapping membership)', async t => {
+  // there is no preferred epoch, you will need to choose choose a winner
+  // and exclude some members
+  // (not sure what the API should be here)
```
Powersource, line 336: maybe, return an epoch that the user can use for now (in case they're planning to post a message this second), but also include a soft error that stuff needs to be resolved asap?

test/lib/epochs.test.js
mixmix: This is a test file. I am manually getting what I think is the solution to test against the function. Clunky but fine as it's a controlled setup

PR #104, opened 2023-05-03
refactor epochs to consolidate getPreferredEpoch

mixmix:
merges into this pr https://github.com/ssbc/ssb-tribes2/pull/95

This PR merges pickEpoch and getPreferredEpoch

Makes changes to make many of the epoch functions stream-friendly 

* [x] update readme
* [x] add safety check to `publish` that makes sure the picked epoch is the preferred epoch (feels a bit weird? but good enough for now?)


index.js
```diff
@@ -319,19 +321,15 @@ module.exports = {
...
         pull.flatten(),
-        pull.map((update) => {
-          if (update.added) return update.added
-          return update
-        }),
+
         // if it's not live, only include adds
-        pull.filter((update) => opts?.live || !update.excluded)
+        live ? null : pull.filter((update) => update.added)
       )
```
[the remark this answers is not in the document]
mixmix: TODO : discuss if that's fine, someone update the README if needed

index.js
```diff
@@ -319,19 +321,15 @@ module.exports = {
...
+        pull.map((epoch) => getMembers.stream(epoch.id, { live })),
         pull.flatten(),
-        pull.map((update) => {
-          if (update.added) return update.added
-          return update
-        }),
+
         // if it's not live, only include adds
-        pull.filter((update) => opts?.live || !update.excluded)
+        live ? null : pull.filter((update) => update.added)
```
Powersource, line 332: not sure why i added this line to start with, if it's not live then it's not going to output any exclusions anyway (if someone's excluded then we'd switch to a new epoch without exclusions on it)

Powersource (changes requested):
test failing!

lib/epochs.js
```diff
@@ -267,119 +278,102 @@ function reduceEpochs(ssb, groupId, getters = {}, cb) {
...
-    pull.map((msg) => {
-      const { type, recps, excludes } = msg.value.content
-      if (type === 'group/add-member')
-        return recps.slice(1).map((feedId) => ({ added: feedId }))
-      else return excludes.map((feedId) => ({ excluded: feedId }))
-    }),
-    pull.flatten(),
-    // grab updates until we get an exclusion, then close the stream (even if live)
-    // last: true includes the exclusion as well right before closing
-    pull.take((update) => update.added, { last: true })
```
Powersource, line 356: i see you moved this function to `getMembers.stream`, that's fine

but i think this line was important, it was part of a bugfix. i suspect the last listMembers test will fail without it

index.js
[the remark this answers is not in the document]
Powersource: added todo about readme.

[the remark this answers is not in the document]
Powersource: it feels weird for the un-live version but maybe it's the best compromise that works both live and unlive

index.js
```diff
@@ -319,19 +321,15 @@ module.exports = {
...
     function listMembers(groupId, opts = {}) {
+      const { live } = opts
+
       return pull(
-        getPickedEpoch(groupId, { live: !!opts?.live }),
-        pull.map((pickedEpoch) => {
-          console.log('pickedEPoch', pickedEpoch)
-          return getMemberUpdates(pickedEpoch.id, { live: !!opts?.live })
-        }),
+        getPreferredEpoch.stream(groupId, { live }),
```
Powersource, line 327: i guess my worry about not having `getPickedEpoch` is that, since we use the picked key when posting, but then use `getPreferredEpoch` for everything else, there can in theory be race conditions where a new epoch is preferred but we haven't had time to pick it.

that's why i was thinking `getPreferredEpoch` fitted better as a thing in the background. it's fine if it's slow/hasn't caught up to everything yet, since everything related to posting/writing is all based on the picked key.

lib/epochs.js
mixmix: true enough! I found another solution but I think this is probably more elegant ... lemme look

index.js
mixmix: yes the question is "what is the source of truth" ?

I hadn't fully understood what you were going for with picked vs preferred. perhaps we could check if the preferredEpoch is in the keystore before using it.

index.js
```diff
@@ -319,28 +322,42 @@ module.exports = {
...
-        pull.map((update) => {
-          if (update.added) return update.added
-          return update
-        }),
-        // if it's not live, only include adds
-        pull.filter((update) => opts?.live || !update.excluded)
+        getPreferredEpoch.stream(groupId, { live }),
+        pull.map((epoch) => getMembers.stream(epoch.id, { live })),
+        pullFlatMerge()
+        // NOTE pullFlatMerge needed in live case otherwise new streams don't get a word in!
```
mixmix, line 351: this was my solution to the "oh no the last strep didn't close" thing
Its sucky but I think we need to leave the streams open to keep hearing the exclusions.

I kinda hate we have this live stream... it has been so much work to get sorted and I'm still not confident with how it will behave in edge cases : 
- what happens when there are multiple exclusions happen for a epoch (you an I fork the group) ... this live stream should do what?
- in the case of a fork ... what members am I getting?

Safest thing to do would be to make this not live IMO.
Until we need live streaming members. In my experience very few things need live streaming, and even then I tend to use the live stream more as a trigger to run some async process. MUCH simpler.

What do you think?

lib/epochs.js
mixmix: https://github.com/ssbc/ssb-tribes2/pull/104#discussion_r1183642865

p.s. I didn't know you could use pull.take like this. Awesome

index.js
mixmix: I think I agree with you. I think I disagree that it's something for `lib/epochs` because you're not calculating anything with epochs, you're just using the latest write keys? But maybe I don't have the full overview for how you were using this.

Keen to discuss tonight

index.js
Powersource: We need livestreaming for replication scheduler https://github.com/ssbc/ssb-replication-scheduler/blob/2c5133150ae680c60981e4e2f783365ee83a95a2/index.js#L82

Powersource: > what happens when there are multiple exclusions happen for a epoch (you an I fork the group) ... this live stream should do what?

instead of emitting smaller chunks of state updates (like we do now) maybe it's simpler to go with my previous suggestion of `Stream<ListOfAllCurrentMembers[]>` where it's very clear what the reduced state is?

index.js
Powersource: > perhaps we could check if the preferredEpoch is in the keystore before using it.

we have `publish()` where we could add a check but we also use db.create in tribes2.create... although that is once, and at that point which key to use is really obvious

Powersource: added todo at the top

Powersource (changes requested):
1. do the todos (added in the top comment)
2. maybe change the stream format (see open thread)

Powersource:
@mixmix boop

mixmix:
Cool, will execute these changes then merge

index.js
mixmix: Good idea - I like this much better

index.js
```diff
@@ -319,33 +324,59 @@ module.exports = {
...
+            })
+          })
+          return
+        }
+
+        let abortable = pullAbortable()
+        const source = pull(
+          getPreferredEpoch.stream(groupId, { live }),
+          pull.map((epoch) => {
+            abortable.abort()
```
mixmix, line 356: added killing the old stream after we finds a new epoch

index.js
```diff
@@ -466,33 +498,36 @@ module.exports = {
...
         // prettier-ignore
         if (err) return cb(clarify(err, 'Error finding or creating additions feed when starting ssb-tribes2'))
-        return cb()
+        return cb(null)
       })
 
       ssb.metafeeds.findOrCreate((err, myRoot) => {
         // prettier-ignore
-        if (err) return cb(clarify(err, 'Error getting own root in start()'))
+        if (err) return console.log(clarify(err, 'Error getting own root in start()'))
```
mixmix, line 506: in `start()` you had multiple places you were calling cb.
I changed it so that cb is only called once and subsequent errors are just logged.

index.js
```diff
@@ -527,33 +562,88 @@ module.exports = {
...
             },
             (err) => {
               // prettier-ignore
-              if (err) return cb(clarify(err, "Error finding new epochs we've been added to"))
+              if (err) return console.error(clarify(err, "Error finding new epochs we've been added to"))
             }
           )
         )
+
+        // listen for new epochs and update groupInfo as required
```
mixmix, line 593: added this new listener for when you discover a new group epoch or are added to a group.
TODO - this needs merging with the listener above it I think, but I didn't have time to do this

lib/epochs.js
```diff
@@ -99,141 +101,196 @@ module.exports = function Epochs(ssb) {
...
+        cb(null, {
+          added: [...added],
+          toExclude: [...toExclude],
         })
-      )
-    },
+      })
+    )
+  }
+  getMembers.stream = function getMembersStream(epochRootId, opts = {}) {
```
mixmix, line 130: changed `getMembers.stream` so that it consolidates updates emitted:
- loads current state and emits that
- only emits when there is new info discovered (and then emits the whole current state again)

Very happy with how this turned out. Wrote some tests over it too

index.js
Powersource: woa this was confusing at first but ok you sort of reuse the same aborter. and you don't want getMembers to necessarily end automatically so you do the aborting here

index.js
Powersource: mm smart. but do we want something more self-destructive than a console.log? (also afraid the logs will be caught in the linter lol)

mixmix: I don't know, but `cb` was not an option!

PR #112, opened 2023-05-11
preferredEpoch (live) now emits on membership updates

mixmix:
Found an intermittently failing test that needed fixing

### DIFF
--- a/lib/epochs.js
+++ b/lib/epochs.js
@@ -6,11 +6,13 @@ const { promisify: p } = require('util')
 const { fromMessageSigil } = require('ssb-uri2')
 const pull = require('pull-stream')
 const pullDefer = require('pull-defer')
+const pullFlatMerge = require('pull-flat-merge')
 const Reduce = require('@tangle/reduce')
 const OverwriteFields = require('@tangle/overwrite-fields')
 const clarify = require('clarify-error')
 const Butt64 = require('butt64')
 const isCanonicalBase64 = require('is-canonical-base64')
+const { where, and, type, live, toPullStream } = require('ssb-db2/operators')
 const {
   validator: {
     group: {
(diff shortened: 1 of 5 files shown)


lib/epochs.js
```diff
@@ -286,6 +295,93 @@ module.exports = function Epochs(ssb) {
...
+      })
+    )
+  }
+  // emits updated @tangle/reduce objects as changes occur to epoch graph
+  epochsReduce.stream = function epochsReduceStream(groupId, opts = {}) {
+    const { getters, live: isLive } = opts
+    const reduce = new Reduce(strategy)
+
+    return pull(
+      // reduce can be modified by 2 sorts of update:
```
mixmix, line 341: The problem was here - `epochsReduceStream` was not emitting new updates when epochs changed in ways that count. Formally the epoch tangle would be emitted ONLY if a new epoch was discovered... and the membership of that epoch would be fixed in the tangle data at that point - regardless of whether new `add-member` messages came in.

This is a problem for calculating `preferredEpoch` - as that consumes this data, and assumes the tangle has all the info it needs.

This updates the epoch tangle graph as new people are added to the graph.
(could have done the same with `exclude-member` but will defer till we need that live updating)

PHEW

test/lib/epochs.test.js
```diff
@@ -313,14 +310,15 @@ test('lib/epochs (getPreferredEpoch - 4.4. same membership)', async (t) => {
...
   pull(
     Epochs(oscar).getPreferredEpoch.stream(group.id, { live: true }),
     pull.drain(
-      (epoch) => {
-        livePreferredEpoch = epoch
-      },
-      (err) => {}
+      (epoch) => livePreferredEpochs.unshift(epoch),
+      (err) => {
+        if (err && testsRunning) t.error(err, 'getPreferredEpoch.stream smooth')
```
mixmix, line 320: because there was no error handling on the drain, errors were getting swallowed :cry: which wasted a lot of time.
This is an anti-pattern of drain IMO - I think it would be smarter to require an error handling function even if it's `() => {}`

Powersource: well there was one but it was `(err)=> {}`

PR #116, opened 2023-05-17
Replicate!

mixmix:
Sidestep replication problems by allowing a fallback to "legacy" replication.

### DIFF
--- a/test/add-member.test.js
+++ b/test/add-member.test.js
@@ -6,10 +6,11 @@ const test = require('tape')
 const pull = require('pull-stream')
 const { promisify: p } = require('util')
 const ssbKeys = require('ssb-keys')
-const Testbot = require('./helpers/testbot')
-const replicate = require('./helpers/replicate')
+const { Testbot, replicate, Run } = require('./helpers')
 
 test('get added to a group', async (t) => {
+  const run = Run(t)
+
   const alice = Testbot({
     keys: ssbKeys.generate(null, 'alice'),
     mfSeed: Buffer.from(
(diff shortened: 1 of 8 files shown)


test/add-member.test.js
```diff
@@ -25,64 +26,42 @@ test('get added to a group', async (t) => {
...
+    'tribes2 started for both alice and bob',
+    Promise.all([alice.tribes2.start(), bob.tribes2.start()])
+  )
 
   await p(alice.metafeeds.findOrCreate)()
   const bobRoot = await p(bob.metafeeds.findOrCreate)()
 
-  await replicate(alice, bob)
-  t.pass('alice and bob replicate their trees')
+  await run('alice and bob replicate their trees', replicate(alice, bob))
```
mixmix, line 37: I added this run function because it tidies up the tests and also gives much better logging for free.
relevant because I was debugging some replicate errors

test/helpers/replicate.js
```diff
@@ -6,29 +6,50 @@ const { promisify: p } = require('util')
...
+  )
+
+  const isSync = await ebtReplicate(person1, person2).catch((err) =>
+    console.error('Error with ebtReplicate:\n', err)
+  )
+
+  await p(conn.close)(true).catch(console.error)
+
+  if (!isSync) {
+    console.error('EBT failed to replicate! Final state:')
```
mixmix, line 27: Put some verbose logging in so we can see exactly what went wrong with EBT replication.

test/helpers/replicate.js
```diff
@@ -6,29 +6,50 @@ const { promisify: p } = require('util')
...
+    console.log(person1.id, await p(person1.getVectorClock)())
+    console.log(person2.id, await p(person2.getVectorClock)())
 
+    console.error('falling back to legacyReplicate')
+    await legacyReplicate(person1, person2)
```
mixmix, line 32: Added a fallback to legacy replication.
This function is different in that is pulls the metafeeds from person1 to person2 and vice versa (i.e. doesn't technically put the databases in the same state, as other peoples feeds will not be synced by this.)

mixmix: this didn't work out. Could make it work but it's not straight fwd

mixmix:
changed this PR into [another one](https://github.com/ssbc/ssb-tribes2/pull/117), closing

PR #106, opened 2023-05-04
People added late to a group should get access to all epochs

Powersource:
Fixes https://github.com/ssbc/ssb-tribes2/issues/76

* [ ] add-member people to all tip epochs
* [ ] send added people all old secrets as well
* [ ] when joining a group, grab all tip invites (maybe we do already?) and all the secrets in them
* [ ] test that we add to all epoch tips?
* [ ] test that a late added member can read old stuff

PR #117, opened 2023-05-18
added-late_CI_FIXES

mixmix:
Wanted to get this passing reliably. Also changes `replicate` to "replicate all" by default.

The function I wrote goes through each pair of peers in a set (with repetition). Turns our the repetition doesn't cost anything to do, because the vector clock check is _so dang fast_

```
alice-bob   ▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨ 3279ms
alice-carol ▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨ 4147ms
alice-david ▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨ 5414ms
bob-alice   ▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨ 5385ms
bob-carol   ▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨ 5374ms
bob-david   ▨ 28ms
carol-alice ▨ 11ms
carol-bob   ▨ 6ms
carol-david ▨ 6ms
david-alice ▨ 5ms
david-bob   ▨ 5ms
david-carol ▨ 8ms
```
- see `alice-david` is slow the first time, but `david-alice` later is almost instantaneous.
- interestingly `alice-bob` and `bob-alice` are both slow ... this is because in between these alice has picked up more feeds from carol and david (I think?)

### DIFF
--- a/test/helpers/replicate.js
+++ b/test/helpers/replicate.js
@@ -7,10 +7,38 @@ const pull = require('pull-stream')
 const pullMany = require('pull-many')
 const deepEqual = require('fast-deep-equal')
 
+module.exports = async function replicate(...peers) {
+  if (peers.length === 1 && Array.isArray(peers[0])) peers = peers[0]
+  if (peers.length === 2) return replicatePair(...peers)
+
+  return pull(
+    pull.values(peers),
+    pull.asyncMap((person1, cb) => {
+      pull(
+        pull.values(peers),
+        pull.asyncMap((person2, cb) => {
+          if (person1.id === person2.id) return cb(null, true)
+
+          replicatePair(person1, person2)
+            .then((res) => cb(null, true))
+            .catch((err) => cb(err))
+        }),
+        pull.collect(cb)
+      )
+    }),
+    pull.collectAsPromise()
+  )
+}
+
 /**
  * Fully replicates person1's metafeed tree to person2 and vice versa
  */
-module.exports = async function replicate(person1, person2) {
+async function replicatePair(person1, person2) {
+  // const start = Date.now()
+  // const ID = [person1, person2]
+  //   .map(p => p.name || p.id.slice(0, 10))
+  //   .join('-')
+
   // Establish a network connection
   const conn = await p(person1.connect)(person2.getAddress())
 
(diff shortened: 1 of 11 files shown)


test/helpers/replicate.js
```diff
@@ -7,33 +7,73 @@ const pull = require('pull-stream')
...
 
   // Wait until both have replicated all feeds in full (are in sync)
-  let i = 0
-  await retryUntil(async () => {
-    i++
+  const isSync = async () => {
     const clocks = await Promise.all([
       p(person1.getVectorClock)(),
       p(person2.getVectorClock)(),
     ])
```
Powersource, line 70: This replicate function will probably have the same bug that we ran into in https://github.com/ssbc/ssb-tribes2/pull/106#discussion_r1196639404

I think this might break if it's bob that creates the group.

test/helpers/replicate.js
```diff
@@ -45,21 +85,8 @@ function setupFeedRequests(person1, person2) {
     ]),
```
Powersource, line 85: "This is the list of feedIds that we should care about in our vector clock" mix says

Powersource:
added a comment to replicate in the other pr after merging

staltz:
> `alice-bob   ▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨▨ 3279ms`

All I can think of is how cute these boxes are.

mixmix:
haha @staltz I see you're also a fan of command line logging

PR #119, opened 2023-05-19
Add allAdded arg to listMembers

Powersource:
* [ ] Update readme

### DIFF
--- a/index.js
+++ b/index.js
@@ -24,6 +24,7 @@ const {
       addMember: isAddMember,
       content: isContent,
       excludeMember: isExcludeMember,
+      initRoot: isInitRoot,
       initEpoch: isInitEpoch,
     },
   },
@@ -39,6 +40,7 @@ const publishAndPrune = require('./lib/prune-publish')
 const MetaFeedHelpers = require('./lib/meta-feed-helpers')
 const Epochs = require('./lib/epochs')
 const { groupRecp } = require('./lib/operators')
+const { slowPredicate } = require('jitdb/operators')
 
 module.exports = {
   name: 'tribes2',
@@ -359,7 +361,7 @@ module.exports = {
     }
 
     function listMembers(groupId, opts = {}) {
-      const { live } = opts
+      const { live, allAdded } = opts
       const deferredSource = pullDefer.source()
 
       get(groupId, (err, group) => {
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -368,6 +370,36 @@ module.exports = {
         // prettier-ignore
         if (group.excluded) return deferredSource.abort( new Error("We're excluded from this group, can't list members"))
 
+        if (allAdded) {
+          // TODO: make secrets live. box2 live get?
```
Powersource, line 374: make more streamy

index.js
```diff
@@ -368,6 +371,65 @@ module.exports = {
...
+          pull(
+            ssb.db.query(
+              where(
+                and(
+                  isDecrypted('box2'),
+                  type('group/add-member'),
+                  groupRecp(groupId)
+                )
+              ),
+              dbLive({ old: true }),
```
Powersource, line 389: this needed to be live for some reason :thinking:  but if we're in the group and have the group id then this msg should've already been decrypted, it's where we found out about the group to begin with

index.js
```diff
@@ -368,6 +371,67 @@ module.exports = {
...
+                    )
+                  ),
+                  live ? dbLive({ old: true }) : null,
+                  toPullStream()
+                ),
+                pull.filter((msg) => isInitRoot(msg) || isInitEpoch(msg)),
+                pull.map((msg) => msg.key),
+                pull.unique(),
+                pull.map((epochRootId) =>
+                  getMembers.stream(epochRootId, { live })
```
Powersource, line 421: we need to adapt this function to getMembers having this `{added, toExclude}` format

index.js
Powersource: didn't need secrets here

index.js
Powersource: simplified away this

index.js
```diff
@@ -368,6 +371,67 @@ module.exports = {
...
+
+              const source = pull(
+                ssb.db.query(
+                  where(
+                    and(
+                      type('group/init'),
+                      slowPredicate(
+                        'secret',
+                        (secret) => secrets.includes(secret),
+                        { name: secrets.join('') }
```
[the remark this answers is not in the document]
Powersource: didn't need this

Powersource:
not testing this arg here but it is tested in https://github.com/ssbc/ssb-replication-scheduler/pull/20

Powersource:
@mixmix should be done. i'll just do the readme tomorrow

mixmix (approved):
Looks good to me!

The only comment I have is "we are getting some real fucken big functions! I would prefer to either make a standalone method `listAllMembers` or even just an internal function called that so that `listMembers` is not a behemoth that is hard to read and maintain. (my thinking is the more lines of code there are in a function, the more likely it is for point mutations to cause bad bugs)

Powersource:
1.2.0

PR #120, opened 2023-05-22
replicate: don't try to sync feeds you can't decrypt

mixmix:


Powersource (changes requested):
nice, and cool that we didn't need to update any tests

only request is to remove this comment that i added before that warned of this. https://github.com/ssbc/ssb-tribes2/blob/ab0926171ca4ca8edc39da435bc38391e47b5116/test/helpers/replicate.js#L12

 (and maybe verify by trying out putting bob first in these args https://github.com/ssbc/ssb-tribes2/blob/ab0926171ca4ca8edc39da435bc38391e47b5116/test/add-member.test.js#L292 )

mixmix:
Good suggestions. Executed.

Passes locally.
Merging on CI pass

Issue #113, opened 2023-05-11
lib/epochs (getPreferredEpoch - 4.5. subset membership)


Powersource:
https://github.com/ssbc/ssb-tribes2/blob/e3591f6620a3e8ecd50e693579d604d4aa65c147/test/lib/epochs.test.js#L321C12-L325

PR #122, opened 2023-05-24
getPreferredEpoch case 4.5 - subset membership

mixmix:
fixes https://github.com/ssbc/ssb-tribes2/issues/113

### DIFF
--- a/lib/epochs.js
+++ b/lib/epochs.js
@@ -24,6 +24,7 @@ const {
   },
 } = require('private-group-spec')
 const difference = require('set.prototype.difference')
+const isSubsetOf = require('set.prototype.issubsetof')
 
 const { groupRecp } = require('./operators')
 const getTangleUpdates = require('./tangles/get-tangle-updates')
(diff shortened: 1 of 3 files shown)


lib/epochs.js
```diff
@@ -544,13 +545,18 @@ function buildPreferredEpoch(reduce, cb) {
...
-      return cb(
-        Error('Tips do not have same membership - case not handled yet')
-      )
+    } else if (isSubsetOf(members1, members2)) {
+      preferredEpoch = tips[0]
+    } else if (
+      isSubsetOf(members2, members1)
+    ) {
+      preferredEpoch = tips[1]
     }
```
Powersource, line 557: could be nice with some comments here or?

Powersource (approved):
wow that's simple and clean.

clarifying comments could be nice

lib/epochs.js
mixmix: done

Powersource: mm where?

Issue #123, opened 2023-05-24
Crash resistance for excludeMembers


Powersource:
For exclusion we post 3 different messages

1. First the `group/exclude-member` message. We don't need to recover from this, if we crash on this step, the user can tell and they just have to try again.
2. Then a `group/init` to init the new epoch. Hopefully it's enough to look for a 1. msg. But hmm when should we search for that? If we call excludeMembers again with the exact same args? Should excludeMembers maybe just post exclude-member, and msgs 2. and 3. should be left to listeners?
3. Lastly all remaining members are re-added with `group/add-member` messages. The lib/epoch function `getMissingMembers` is probably very helpful here.

Todos:
* [x] Recover after only part of members got re-added https://github.com/ssbc/ssb-tribes2/pull/124
* [x] Recover when no members were re-added https://github.com/ssbc/ssb-tribes2/pull/126
* [x] https://github.com/ssbc/ssb-tribes2/pull/129
* [x] Recover when new epoch init didn't get posted https://github.com/ssbc/ssb-tribes2/pull/127
* [x] https://github.com/ssbc/ssb-tribes2/pull/134
* [x] Recovering someone else should happen after a random timeout
* [x] Disable recovery by default and add config option to enable https://github.com/ssbc/ssb-tribes2/pull/139
* [x] Test that someone else can recover if we don't
* [x] ~Test that we recover on restart of our client~
* [x] ~Maybe also run the recovery functions at the start of all functions that get from/operate on members tangles, e.g. if the user notices that exclusion failed, so they click Exclude again~ hmm the time delays are core in the recovery (not doing it when not needed), and we can't have such delays before running most functions
* [ ] Update tribes2 in the demo once the above PRs are merged https://github.com/ssbc/ssb-tribes2-demo/pull/12

Powersource:
Gonna write some more notes to try to figure out exactly how I should go about this.

Goals:
1. We don't want a dangling exclude-member message in an epoch, that didn't end up having any effect
2. We don't want a dangling epoch init, in an epoch that didn't end up with any members.
3. We everyone to get added to the new epoch except for the excluded peers.

Questions:
1. Do we want to help others with their failed exclusions? Maybe yeah? Since the group/epoch is a common resource and one person crashing might break the group for all of us.
2. When do we fix a broken state? When calling the function again? If other people should be able to fix it too, then they'll want a listener. Do we want to use that listener for ourselves as well?

Powersource:
[...]

Issue #107, opened 2023-05-04
In crash-resistance tests, use secret opts that make the functions crash at certain points


Powersource:
Instead of trying to replicate how things probably look after a crash

for instance in `create.test.js` we have hacky code that partially creates a group.

mix suggested at some point (and i think this is probably better now that i've run into some weird test errors due to mismatch (due to updates of the function) between how the function and the emulation works) to use secret opts. something like

```js
create(opts, cb) {
...
createGroupFeed((err, feed) => {
  if (err || opts._crashOnCreateGroupFeed) return cb(err)
```

PR #124, opened 2023-05-24
Re-add missing members if only some were re-added after failed excludeMembers

Powersource:
Part of https://github.com/ssbc/ssb-tribes2/issues/123

* [x] Test. Use this trick https://github.com/ssbc/ssb-tribes2/issues/107
* [x] document timeoutScale


listeners.js
```diff
@@ -192,5 +192,29 @@ module.exports = function startListeners(ssb, onError) {
...
+    pull(
+      ssb.db.query(
+        where(type('group/init')),
+        live({ old: true }),
+        toPullStream()
+      ),
+      pull.filter((msg) => isInitEpoch(msg)),
+      pull.drain(
+        (msg) => {
+          // TODO: add random timeout
```
Powersource, line 207: @mixmix note that I think both you and I want to add "slight random timeout" around now. i think you might get to it first but in that case put it somewhere i can easily reuse maybe? :)

Powersource: for 4.6 i think.

could also be nice with a config that lets us put it to 0 in tests

listeners.js
```diff
@@ -192,5 +192,29 @@ module.exports = function startListeners(ssb, onError) {
...
+
+          // TODO: check if it's the preferred epoch. maybe do that live instead of the query?
+
+          console.log("TODO: re-add people if they're missing")
```
Powersource, line 211: what do we do if we fail with re-adding anyone at all? should we run this at an interval as well?

Powersource: should we run a `checkAndFixMemberTangleIntegrity()` at the start of every function that reads or writes to the member tangle?

README.md
```diff
@@ -194,6 +194,10 @@ Makes sure that you're set up to send and receive group invites, by creating an
 
 - `cb` _Function_ - a callback of signature `(err)`
 
+## Config
+
+You can set the secret stack config `config.tribes2.timeoutScale` to control how slowly the client should try to fix a conflicting state, where other clients might be trying to fix the same conflict at the same time. By default `1000`, which gives a timeout between 1s-30s. A higher value reduces the risk of creating new conflicts since other clients don't do the same conflict resolution at the same time, but increase the time that the group is in an unstable state. A lower number corrects things faster but increases the risk of ending up in new conflicts.
```
Powersource, line 199: Too verbose?

listeners.js
Powersource: did this here now https://github.com/ssbc/ssb-tribes2/blob/bdf1a2ee676c8ff27a99a5c22bd156f07a51d4ff/listeners.js#L211-L222

should be easily extractable to a helper function

listeners.js
Powersource: unresolved question but i'm thinking it'll be for another pr

listeners.js
```diff
@@ -192,5 +198,43 @@ module.exports = function startListeners(ssb, onError) {
...
+          pull.drain(
+            () => {
+              const timeoutScale = config.tribes2?.timeoutScale ?? 1000
+              const timeoutRandom = Math.random() * 30 + 1
+              const randomTimeout = timeoutScale * timeoutRandom
+
+              const timeoutId = setTimeout(() => {
+                reAddMembers(ssb, group.id, null, (err) => {
+                  // prettier-ignore
+                  if (err && !isClosed) return onError(clarify(err, 'todo'))
```
Powersource, line 218: oh yeah i gotta clarify the errs

listeners.js
Powersource: testing this here https://github.com/ssbc/ssb-tribes2/pull/126 and also added a task in the issue for running checks at the start of member functions

listeners.js
Powersource: extracting this in https://github.com/ssbc/ssb-tribes2/pull/127

README.md
mixmix: no, perfect.
I think your code technically results in 1-31s default , but I don't care <3

test/exclude-members.test.js
```diff
@@ -684,3 +685,94 @@ test("restarting the client doesn't make us rejoin old stuff", async (t) => {
   await p(alice.close)(true)
   await p(bob.close)(true)
 })
+
+test('On exclusion, if we fail to re-add all people, someone else does that instead', async (t) => {
+  const run = Run(t)
+  const alice = Testbot({ name: 'alice', timeoutScale: 300 * 1000 })
```
mixmix, line 691:
```suggestion
  const alice = Testbot({ name: 'alice', timeoutScale: 300 * 1000 })
  // set alice to be slow to fix mistakes, to allow carol time to do it
```

test/exclude-members.test.js
```diff
@@ -684,3 +685,94 @@ test("restarting the client doesn't make us rejoin old stuff", async (t) => {
...
+    'alice gets her complete list',
+    pull(alice.tribes2.listMembers(groupId), pull.collectAsPromise())
+  )
+
+  t.deepEqual(
+    aliceNewList[0].added.sort(),
+    [aliceId, carolId, davidId].sort(),
+    'group is also fixed for alice'
+  )
+
```
mixmix, line 771: I am being paranoid, but I think you should check david , carol, alice are all agreed on the preferredEpoch, or posting with the same encrypted key (and it's not the initial one, i.e. bob)

As an aside, I encourage you to make "oscar" the one who is default excluded in tests. Oscar the Grouch ... or Opposition. It makes the tests marginally easier to parse when we have a familiar cast with given roles

listeners.js
```diff
@@ -192,5 +198,43 @@ module.exports = function startListeners(ssb, onError) {
...
       )
     )
+
+    // re-add missing people to a new epoch if the epoch creator didn't add everyone but they added us.
+    // we're only doing this for the preferred epoch atm
+    pull(
+      ssb.tribes2.list({ live: true }),
+      pull.map((group) =>
+        pull(
+          getPreferredEpoch.stream(group.id, { live: true }),
```
mixmix, line 208: This works, but I worry it means opening N listening streams for N groups, and `getPreferredEpoch` live is a lot.
I think we can replace all this with one listener for `group/init` messages aye? then pull the `groupId` out of the recps? 

could be a TODO / issue as this is working

mixmix (approved):
Great work @Powersource !

Couple of questions, but I think those can be turned into other issues if we want

mixmix:
Seems like failing test is on me : 
image

Happily alice and bob agree on the preferred epoch, but might both be off-spec D:

test/exclude-members.test.js
Powersource: idk if checking this is that important in this test? or how do you imagine this could go wrong?

lib/exclude.js
```diff
@@ -0,0 +1,51 @@
...
+    if (err) return cb(clarify(err, "Couldn't get preferred epoch when re-adding members"))
+
+    epochs.getMissingMembers(groupId, (err, missingMembersEveryEpoch) => {
+      // prettier-ignore
+      if (err) return cb(clarify(err, "Couldn't find missing members after exclusion"))
+
+      const missingMembers = missingMembersEveryEpoch
+        .find((epoch) => epoch.epoch === preferredEpoch.id)
+        ?.missing.filter((member) =>
+          opts?._reAddSkipMember ? member !== opts?._reAddSkipMember : true
```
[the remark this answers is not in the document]
Powersource: i think prettier would remove those parens

lib/exclude.js
```diff
@@ -0,0 +1,51 @@
...
+        )
+
+      if (!missingMembers) return cb()
```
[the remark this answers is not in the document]
Powersource: you mean null as in an error or a lack of one? this isn't an error but more of a NOP

test/exclude-members.test.js
Powersource: added, but on the line before!

lib/exclude.js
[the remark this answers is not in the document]
Powersource: yeah it does. but refactored out the function.

lib/exclude.js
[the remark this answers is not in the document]
Powersource: changed the if

listeners.js
```diff
@@ -192,5 +198,43 @@ module.exports = function startListeners(ssb, onError) {
...
+          pull.drain(
+            () => {
+              const timeoutScale = config.tribes2?.timeoutScale ?? 1000
+              const timeoutRandom = Math.random() * 30 + 1
```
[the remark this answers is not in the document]
Powersource: changed to 5s-30s

listeners.js
Powersource: i think i did this partially because i wanted to check if the epoch was preferred anyway, so felt like i could do it this way to both get the epoch and check at the same time

Powersource:
wait where did you get the failing test?

Issue #30, opened 2022-11-03
Use npm module `debug`


staltz:
Like ssb-replication-scheduler uses, also jitdb uses it.

mixmix:
e.g. https://github.com/ssbc/ssb-replication-scheduler/blob/8550b7e5b26429453bb6d3f05f900f8315f4fc4c/metafeed-finder.js#L6

PR #126, opened 2023-05-30
Recover on creating a new epoch but failing to re-add anyone

Powersource:
Based on https://github.com/ssbc/ssb-tribes2/pull/124

Progresses https://github.com/ssbc/ssb-tribes2/issues/123

PR #127, opened 2023-05-30
Recover by creating a new epoch if someone only posted an exclude message

Powersource:
Based on https://github.com/ssbc/ssb-tribes2/pull/126

Progresses https://github.com/ssbc/ssb-tribes2/issues/123

Fixes https://github.com/ssbc/ssb-tribes2/issues/30

### DIFF
--- a/index.js
+++ b/index.js
@@ -269,6 +269,10 @@ module.exports = {
                 tangles: ['epoch'],
                 isValid: isInitEpoch,
               }
+
+              // prettier-ignore
+              if (opts?._newEpochCrash) return cb(new Error('Intentional crash before creating new epoch'))
+
               publish(newEpochContent, newTangleOpts, (err) => {
                 // prettier-ignore
                 if (err) return cb(clarify(err, "Couldn't post init msg on new epoch when excluding members"))
--- a/index.js
+++ b/index.js
@@ -24,12 +24,10 @@ const {
       addMember: isAddMember,
       content: isContent,
       excludeMember: isExcludeMember,
-      initEpoch: isInitEpoch,
     },
   },
   keySchemes,
 } = require('private-group-spec')
-const { SecretKey } = require('ssb-private-group-keys')
 const { fromMessageSigil, isBendyButtV1FeedSSBURI } = require('ssb-uri2')
 
 const startListeners = require('./listeners')
@@ -39,7 +37,7 @@ const publishAndPrune = require('./lib/prune-publish')
 const MetaFeedHelpers = require('./lib/meta-feed-helpers')
 const Epochs = require('./lib/epochs')
 const { groupRecp } = require('./lib/operators')
-const { reAddMembers } = require('./lib/exclude')
+const { createNewEpoch, reAddMembers } = require('./lib/exclude')
 
 module.exports = {
   name: 'tribes2',
@@ -224,71 +222,35 @@ module.exports = {
       if (cb === undefined)
         return promisify(excludeMembers)(groupId, feedIds, opts)
 
-      ssb.metafeeds.findOrCreate(function gotRoot(err, myRoot) {
+      const excludeContent = {
+        type: 'group/exclude-member',
+        excludes: feedIds,
+        recps: [groupId],
+      }
+      const excludeOpts = {
+        tangles: ['members'],
+        isValid: isExcludeMember,
+      }
+      publish(excludeContent, excludeOpts, (err) => {
         // prettier-ignore
-        if (err) return cb(clarify(err, "Couldn't get own root when excluding members"))
+        if (err) return cb(clarify(err, 'Failed to publish exclude msg'))
 
-        const excludeContent = {
-          type: 'group/exclude-member',
-          excludes: feedIds,
-          recps: [groupId],
-        }
-        const excludeOpts = {
-          tangles: ['members'],
-          isValid: isExcludeMember,
-        }
-        publish(excludeContent, excludeOpts, (err) => {
+        // prettier-ignore
+        if (opts?._newEpochCrash) return cb(new Error('Intentional crash before creating new epoch'))
+
+        createNewEpoch(ssb, groupId, (err) => {
           // prettier-ignore
-          if (err) return cb(clarify(err, 'Failed to publish exclude msg'))
+          if (err) return cb(clarify(err, "Couldn't create new epoch when excluding members"))
 
-          const newSecret = new SecretKey()
-          const addInfo = { key: newSecret.toBuffer() }
+          // prettier-ignore
+          if (opts?._reAddCrash) return cb(new Error('Intentional crash before re-adding members'))
 
-          ssb.box2.addGroupInfo(groupId, addInfo, (err) => {
-            // prettier-ignore
-            if (err) return cb(clarify(err, "Couldn't store new key when excluding members"))
-
-            const newKey = {
-              key: newSecret.toBuffer(),
-              scheme: keySchemes.private_group,
-            }
-            ssb.box2.pickGroupWriteKey(groupId, newKey, (err) => {
-              // prettier-ignore
-              if (err) return cb(clarify(err, "Couldn't switch to new key for writing when excluding members"))
-
-              const newEpochContent = {
-                type: 'group/init',
-                version: 'v2',
-                secret: newSecret.toString('base64'),
-                tangles: {
-                  members: { root: null, previous: null },
-                },
-                recps: [groupId, myRoot.id],
-              }
-              const newTangleOpts = {
-                tangles: ['epoch'],
-                isValid: isInitEpoch,
-              }
-
-              // prettier-ignore
-              if (opts?._newEpochCrash) return cb(new Error('Intentional crash before creating new epoch'))
-
-              publish(newEpochContent, newTangleOpts, (err) => {
-                // prettier-ignore
-                if (err) return cb(clarify(err, "Couldn't post init msg on new epoch when excluding members"))
-
-                // prettier-ignore
-                if (opts?._reAddCrash) return cb(new Error('Intentional crash before re-adding members'))
-
-                reAddMembers(
-                  ssb,
-                  groupId,
-                  { _reAddSkipMember: opts?._reAddSkipMember },
-                  cb
-                )
-              })
-            })
-          })
+          reAddMembers(
+            ssb,
+            groupId,
+            { _reAddSkipMember: opts?._reAddSkipMember },
+            cb
+          )
         })
       })
     }
--- a/index.js
+++ b/index.js
@@ -37,7 +37,7 @@ const publishAndPrune = require('./lib/prune-publish')
 const MetaFeedHelpers = require('./lib/meta-feed-helpers')
 const Epochs = require('./lib/epochs')
 const { groupRecp } = require('./lib/operators')
-const { createNewEpoch, reAddMembers } = require('./lib/exclude')
+const { createNewEpoch } = require('./lib/exclude')
 
 module.exports = {
   name: 'tribes2',
@@ -238,20 +238,19 @@ module.exports = {
         // prettier-ignore
         if (opts?._newEpochCrash) return cb(new Error('Intentional crash before creating new epoch'))
 
-        createNewEpoch(ssb, groupId, (err) => {
-          // prettier-ignore
-          if (err) return cb(clarify(err, "Couldn't create new epoch when excluding members"))
-
-          // prettier-ignore
-          if (opts?._reAddCrash) return cb(new Error('Intentional crash before re-adding members'))
-
-          reAddMembers(
-            ssb,
-            groupId,
-            { _reAddSkipMember: opts?._reAddSkipMember },
-            cb
-          )
-        })
+        createNewEpoch(
+          ssb,
+          groupId,
+          {
+            _reAddCrash: opts?._reAddCrash,
+            _reAddSkipMember: opts?._reAddSkipMember,
+          },
+          (err) => {
+            // prettier-ignore
+            if (err) return cb(clarify(err, "Couldn't create new epoch and/or re-add members when excluding members"))
+            cb()
+          }
+        )
       })
     }
 
--- a/lib/exclude.js
+++ b/lib/exclude.js
@@ -14,7 +14,7 @@ const {
 } = require('private-group-spec')
 const Epochs = require('./epochs')
 
-function createNewEpoch(ssb, groupId, cb) {
+function createNewEpoch(ssb, groupId, opts, cb) {
   ssb.metafeeds.findOrCreate(function gotRoot(err, myRoot) {
     // prettier-ignore
     if (err) return cb(clarify(err, "Couldn't get own root when excluding members"))
@@ -52,7 +52,19 @@ function createNewEpoch(ssb, groupId, cb) {
           // prettier-ignore
           if (err) return cb(clarify(err, "Couldn't post init msg on new epoch when excluding members"))
 
-          cb()
+          // prettier-ignore
+          if (opts?._reAddCrash) return cb(new Error('Intentional crash before re-adding members'))
+
+          reAddMembers(
+            ssb,
+            groupId,
+            { _reAddSkipMember: opts?._reAddSkipMember },
+            (err) => {
+              // prettier-ignore
+              if (err) return cb(clarify(err, "Couldn't re-add members when creating new epoch"))
+              cb()
+            }
+          )
         })
       })
     })
--- a/lib/exclude.js
+++ b/lib/exclude.js
@@ -96,7 +96,7 @@ function reAddMembers(ssb, groupId, opts, cb) {
 
       if (!missingMembers?.length) return cb()
 
-      debug('about to readd %O', missingMembers)
+      debug('about to re-add %O', missingMembers)
       pull(
         pull.values(chunk(missingMembers, 15)),
         pull.asyncMap((membersToAdd, cb) =>
--- a/lib/exclude.js
+++ b/lib/exclude.js
@@ -4,9 +4,62 @@
 
 const chunk = require('lodash.chunk')
 const pull = require('pull-stream')
+const { SecretKey } = require('ssb-private-group-keys')
 const clarify = require('clarify-error')
+const {
+  validator: {
+    group: { initEpoch: isInitEpoch },
+  },
+  keySchemes,
+} = require('private-group-spec')
 const Epochs = require('./epochs')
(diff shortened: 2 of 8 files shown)


listeners.js
```diff
@@ -237,5 +242,19 @@ module.exports = function startListeners(ssb, config, onError) {
...
+    // if we find an exclude and it's not excluding us but we don't find a new epoch, even after a while, then create a new epoch, since we assume that the excluder crashed or something
+    pull(
+      ssb.db.query(
+        where(and(isDecrypted('box2'), type('group/exclude-member'))),
+        dbLive({ old: true }),
+        toPullStream()
+      ),
+      pull.filter(isExcludeMember),
+      pull.map((msg) => msg.value.content),
+      pull.filter((content) => !content.excluded.includes(myRoot.id))
```
Powersource, line 255: this only works if we assume someone excluding only publishes 1 exclude message when excluding. if they published 2 then one of them would slip through here and we'd try to create a new epoch even if we're excluded.

maybe best solution is instead to stream listMembers and check the exclude list there, since i think that function might look for multiple exclude messages (double check)

listeners.js
```diff
@@ -237,5 +242,75 @@ module.exports = function startListeners(ssb, config, onError) {
         }
       )
     )
+
+    // if we find an exclude and it's not excluding us but we don't find a new epoch, even after a while, then create a new epoch, since we assume that the excluder crashed or something
+    // TODO: combine with the above listener?
```
Powersource, line 247: todo

listeners.js
```diff
@@ -237,5 +242,75 @@ module.exports = function startListeners(ssb, config, onError) {
...
+                  pull.filter((members) => members.toExclude.length),
+                  pull.take(1),
+                  pull.drain(
+                    () => {
+                      const timeout = randomTimeout(config)
+
+                      const timeoutId = setTimeout(() => {
+                        ssb.tribes2.get(group.id, (err, group) => {
+                          // prettier-ignore
+                          if (err && !isClosed) return onError(clarify(err, 'todo'))
```
Powersource, line 269: clarify errs

test/exclude-members.test.js
Powersource, line 1: 'On exclusion, recover if we fail to re-add anyone at all' seems to be flaking a bit on 'There was only 1 re-addition message from alice', it gets 2 instead of 1. Seems to happen more locally

Powersource: i wonder if alice's listener runs so fast that it accidentally creates a new epoch while exclusion is happening, ending up with 2 epochs and therefore 2 re-addition msgs

Powersource: also happened in ci now. but not because of the latest commit, it happened when i stashed that one too

Powersource: also happened in ci now. but not because of the latest commit, it happened when i stashed that one too

Powersource: hmm no both messages seem to belong to the same epoch. could be more race conditions? maybe just a race condition in the reAdd function?

```
ok 10 alice excludes bob but crashes before re-adding herself and carol
readdmember res [
  {
    "previous": "%TC5NW1rAOkiQ/W4YTLhFrh4ASAsI7EcQ8SjOWSZQIRw=.sha256",
    "sequence": 3,
    "author": "@M3OAID01925gO7/l3ZjKbtbmUkXw0gvqRwr5n7hwwPk=.ed25519",
    "timestamp": 1685619076372,
    "hash": "sha256",
    "content": {
      "type": "group/add-member",
      "version": "v2",
      "secret": "47qTvz1zDX5aUBXZPfYx45ORSNXculWk8pUeirK+Hyg=",
      "root": "ssb:message/classic/8ZNj1MuaTVEwULaFtrlzlIKqFLcM6NxmiVODovwoY-0=",
      "creator": "ssb:feed/bendybutt-v1/R7oVP6uSBqSmyC-iQxX67S9Zl2498yxJk4NarqMaBtI=",
      "recps": [
        "ssb:identity/group/98leh3cs6xQmqOVAG90zdMI6WhO04KsaR4kkyds25oU=",
        "ssb:feed/bendybutt-v1/R7oVP6uSBqSmyC-iQxX67S9Zl2498yxJk4NarqMaBtI=",
        "ssb:feed/bendybutt-v1/vNKLsZwOzBfYvqaOeaBQqQ6dPy7HE9ogb3DnzlTaonM="
      ],
      "tangles": {
        "group": {
          "root": "ssb:message/classic/8ZNj1MuaTVEwULaFtrlzlIKqFLcM6NxmiVODovwoY-0=",
          "previous": [
            "ssb:message/classic/mFD-lXW4dmVUoYOL0ybqjI5G8fVAORgNHrvvN1JOuMY="
          ]
        },
        "members": {
          "root": "ssb:message/classic/mFD-lXW4dmVUoYOL0ybqjI5G8fVAORgNHrvvN1JOuMY=",
          "previous": [
            "ssb:message/classic/mFD-lXW4dmVUoYOL0ybqjI5G8fVAORgNHrvvN1JOuMY="
          ]
        }
      }
    },
    "signature": "ybDL5FGpyvTCtTedu0NVe0EiDVXuesmLaS3rQsSZi6WOmWObIBAx6Qtq+o+QndvGMV65s2DDgTd3CcmWraw6Bw==.sig.ed25519"
  }
]
readdmember res [
  {
    "previous": "%6lgUNyUwrYgtu9PPbrmTnlyrbJq0z9paj2oBkcjXSog=.sha256",
    "sequence": 4,
    "author": "@M3OAID01925gO7/l3ZjKbtbmUkXw0gvqRwr5n7hwwPk=.ed25519",
    "timestamp": 1685619076457,
    "hash": "sha256",
    "content": {
      "type": "group/add-member",
      "version": "v2",
      "secret": "47qTvz1zDX5aUBXZPfYx45ORSNXculWk8pUeirK+Hyg=",
      "root": "ssb:message/classic/8ZNj1MuaTVEwULaFtrlzlIKqFLcM6NxmiVODovwoY-0=",
      "creator": "ssb:feed/bendybutt-v1/R7oVP6uSBqSmyC-iQxX67S9Zl2498yxJk4NarqMaBtI=",
      "recps": [
        "ssb:identity/group/98leh3cs6xQmqOVAG90zdMI6WhO04KsaR4kkyds25oU=",
        "ssb:feed/bendybutt-v1/R7oVP6uSBqSmyC-iQxX67S9Zl2498yxJk4NarqMaBtI=",
        "ssb:feed/bendybutt-v1/vNKLsZwOzBfYvqaOeaBQqQ6dPy7HE9ogb3DnzlTaonM="
      ],
      "tangles": {
        "group": {
          "root": "ssb:message/classic/8ZNj1MuaTVEwULaFtrlzlIKqFLcM6NxmiVODovwoY-0=",
          "previous": [
            "ssb:message/classic/6lgUNyUwrYgtu9PPbrmTnlyrbJq0z9paj2oBkcjXSog="
          ]
        },
        "members": {
          "root": "ssb:message/classic/mFD-lXW4dmVUoYOL0ybqjI5G8fVAORgNHrvvN1JOuMY=",
          "previous": [
            "ssb:message/classic/6lgUNyUwrYgtu9PPbrmTnlyrbJq0z9paj2oBkcjXSog="
          ]
        }
      }
    },
    "signature": "JPGP2ppjawHR4TV+s4pVyyfudYNK449ua3G2mhu6Vc9F/yB8G9L6XJBgxIJuTaRuMsGGXv+MztmpDjpzun3ZBg==.sig.ed25519"
  }
]
ok 11 got num of additions after exclude
not ok 12 There was only 1 re-addition message from alice
  ---
    operator: equal
    expected: 1
    actual:   2
    at: Test.<anonymous> (/home/me/prj/ssb/ssb-tribes2/test/exclude-members.test.js:857:5)
```

Powersource: I changed the timeout system a bit now so the default in tests is 50ms.

Powersource: hmm now there's a new one

```
# Can exclude a person in a group with a lot of members
ok 1 tribes2 started
ok 2 created a group
ok 3 invited 10 peers
ok 4 invited 10 peers
ok 5 peers accept invites
ok 6 alice was able to exclude bob in a group with many members
key ssb:message/classic/0Im6p8Re4bMMTmgaRod2E85YCvMpz49WBPHFeE6a8FI= found, but is disconnected from main graph
[...]
key ssb:message/classic/0Im6p8Re4bMMTmgaRod2E85YCvMpz49WBPHFeE6a8FI= found, but is disconnected from main graph
not ok 7 bob is excluded from group
  ---
    operator: deepEqual
    expected: |-
      { id: 'ssb:identity/group/MpdBB51KjaHZ_sFCqIcB1nNE-b9jMkvr6gROxeNflj4=', excluded: true }
    actual: |-
      { id: 'ssb:identity/group/MpdBB51KjaHZ_sFCqIcB1nNE-b9jMkvr6gROxeNflj4=', excluded: undefined }
    at: Test.<anonymous> (/home/me/prj/ssb/ssb-tribes2/test/exclude-members.test.js:600:5)
    stack: |-
      Error: bob is excluded from group
          at Test.assert [as _assert] (/home/me/prj/ssb/ssb-tribes2/node_modules/.pnpm/tape@5.6.3/node_modules/tape/lib/test.js:312:48)
          at Test.bound [as _assert] (/home/me/prj/ssb/ssb-tribes2/node_modules/.pnpm/tape@5.6.3/node_modules/tape/lib/test.js:95:17)
          at Test.tapeDeepEqual (/home/me/prj/ssb/ssb-tribes2/node_modules/.pnpm/tape@5.6.3/node_modules/tape/lib/test.js:553:7)
          at Test.bound [as deepEquals] (/home/me/prj/ssb/ssb-tribes2/node_modules/.pnpm/tape@5.6.3/node_modules/tape/lib/test.js:95:17)
          at Test.<anonymous> (/home/me/prj/ssb/ssb-tribes2/test/exclude-members.test.js:600:5)
          at runMicrotasks (<anonymous>)
          at processTicksAndRejections (node:internal/process/task_queues:96:5)
  ...
not ok 8 Others got excluded from the group
  ---
    operator: fail
    at: <anonymous> (/home/me/prj/ssb/ssb-tribes2/test/exclude-members.test.js:619:20)
    stack: |-
      Error: Others got excluded from the group
          at Test.assert [as _assert] (/home/me/prj/ssb/ssb-tribes2/node_modules/.pnpm/tape@5.6.3/node_modules/tape/lib/test.js:312:48)
          at Test.bound [as _assert] (/home/me/prj/ssb/ssb-tribes2/node_modules/.pnpm/tape@5.6.3/node_modules/tape/lib/test.js:95:17)
          at Test.fail (/home/me/prj/ssb/ssb-tribes2/node_modules/.pnpm/tape@5.6.3/node_modules/tape/lib/test.js:406:7)
          at Test.bound [as fail] (/home/me/prj/ssb/ssb-tribes2/node_modules/.pnpm/tape@5.6.3/node_modules/tape/lib/test.js:95:17)
          at /home/me/prj/ssb/ssb-tribes2/test/exclude-members.test.js:619:20
          at runMicrotasks (<anonymous>)
          at processTicksAndRejections (node:internal/process/task_queues:96:5)
          at async Test.<anonymous> (/home/me/prj/ssb/ssb-tribes2/test/exclude-members.test.js:606:3)
  ...

1..8
# tests 8
# pass  6
# fail  2
```

Powersource: So bob thinks he's not excluded, and the other peers all seem to be creating new epochs, ending up with 11 sometimes...

Powersource: hmm when we ask people to create a new epoch we don't also tell them to re-add members, should do that at least

Powersource: re-adding people in the listener now too but didn't seem to affect this test

index.js
```diff
@@ -224,68 +222,35 @@ module.exports = {
...
+        if (err) return cb(clarify(err, 'Failed to publish exclude msg'))
 
-          const newSecret = new SecretKey()
-          const addInfo = { key: newSecret.toBuffer() }
-
-          ssb.box2.addGroupInfo(groupId, addInfo, (err) => {
+        // prettier-ignore
+        if (opts?._newEpochCrash) return cb(new Error('Intentional crash before creating new epoch'))
+
+        createNewEpoch(
```
mixmix, line 241: BEAUTIFUL

listeners.js
```diff
@@ -199,32 +207,74 @@ module.exports = function startListeners(ssb, config, onError) {
...
-              }, randomTimeout)
-
+              }, timeout)
               closeCalls.push(() => clearTimeout(timeoutId))
+
+              // if we find an exclude and it's not excluding us but we don't find a new epoch, even after a while, then create a new epoch, since we assume that the excluder crashed or something
+              pull(
+                getMembers.stream(preferredEpoch.id, { live: true }),
+                pull.filter((members) => members.toExclude.length),
+                pull.take(1),
```
mixmix, line 235: I don't understand `{ live: true }` and `pull.take(1)` ... surely the live is pointless, as this stream gets killed after it first finds something? Maybe this is good in that it waits till it has at least one state?

You might be right, this just surprised/ confused me?

mixmix: also none of the data is used... you're just using this as a way to listen for a new membership event on the membership of an epoch?
I think this emits an initial state anyway right... so why not just run a setTimeout without all this?

mixmix:
OK, have split this into two PRs to simplify debugging:
1. https://github.com/ssbc/ssb-tribes2/pull/129 : extract createNewEpoch
2. https://github.com/ssbc/ssb-tribes2/pull/130 : crash-epoch-init (less)

listeners.js
Powersource: > surely the live is pointless

yeah i don't remember exactly but i think it might not have been emitting instantly so the live lets it wait some ms

Powersource: the live is very needed since the stream starts as soon as you find the current epoch, but we want to stream until we find an exclude, which might happen now/later/never

Powersource: i think a live db query for exclude messages for that epoch might be equivalent though. but can optimize that later

Powersource:
@mixmix is #130 the same as this PR now?

test/exclude-members.test.js
```diff
@@ -538,7 +539,13 @@ test('Get added to an old epoch but still find newer epochs', async (t) => {
 test('Can exclude a person in a group with a lot of members', async (t) => {
   // NOTE: taking db2 debounces/timeouts out breaks this test, so run default
   // db2 config for stability
-  const _Testbot = (opts) => Testbot({ ...opts, db2: {} })
+  const _Testbot = (opts) =>
+    Testbot({
+      ...opts,
+      db2: {},
+      timeoutLow: 300,
+      timeoutHigh: 300,
```
Powersource, line 547: is this cheating?

listeners.js
Powersource: > so why not just run a setTimeout without all this?

because we're specifically only interested in the event where an exclude was just published, for the preferred epoch

Powersource: i'd argue we can mark this conversation resolved?

test/list-members.test.js
```diff
@@ -162,6 +162,8 @@ test('listMembers works with exclusion', async (t) => {
       '00000000000000000000000000000000000000000000000000000000000ca201',
       'hex'
     ),
+    timeoutLow: 0.5,
+    timeoutHigh: 0.7,
```
Powersource, line 166: just wrote this in chat about this test:

hmm quick hunch in one test:

alice adds bob and carol
alice removes bob
carol creates a new epoch too for some reason
should be fine in theory but alice doesn't get that epoch, and when alice later adds david, it's only to her own epoch
then when alice gets carol's epoch, there's a conflict
alice picks carol's epoch as preferred since it has fewer members

-> wrong member list, missing david

easiest "fix" for all this is just to disable recovery in most tests (set timeout high (ie default)). then selectively enable in a few cases that we actually want to test

i might just also switch from this to doing the demo app. would be fun to show in berlin. and would show us what bugs actually exist IRL

Powersource:
tests passing now (just linter failing). mostly since we disabled recovery in most tests, so, limited knowledge gained. wondering if we should just merge and see how this fares in the demo app. the situations we're testing just get more and more complex :sweat: :shrug:

Powersource:
demo app mostly working now, gonna try this branch out

Powersource:
Doesn't seem to break stuff a lot but there's this bug that pops up (i think unrelated to this PR) that probably stops the code in this PR from running at all https://github.com/ssbc/ssb-tribes2-demo/issues/7

Powersource:
@mixmix see #135 where i (using the demo) tested this PR together with #134 that allows the code in this PR to get run (i think?). doesn't seem to break stuff, so i'd argue we just merge this and maybe polish more later

mixmix:
# Fresh review (in progress)

lib/exclude.js
```diff
@@ -93,6 +96,7 @@ function reAddMembers(ssb, groupId, opts, cb) {
 
       if (!missingMembers?.length) return cb()
 
+      debug('about to readd %O', missingMembers)
```
mixmix, line 99:
```suggestion
      debug('about to re-add %O', missingMembers)
```

test/exclude-members.test.js
```diff
@@ -780,10 +788,10 @@ test('On exclusion, if we fail to re-add all people, someone else does that inst
 
 test('On exclusion, recover if we fail to re-add anyone at all', async (t) => {
   const run = Run(t)
-  const alice = Testbot({ name: 'alice', timeoutScale: 0 })
+  const alice = Testbot({ name: 'alice', ...lowTimeouts })
```
mixmix, line 791: :hot_pepper: - IMO use the API that ssb-tribes uses, so people reading test code don't get confused by API

```suggestion
  const alice = Testbot({ name: 'alice', tribes2: lowTimeouts })
```

test/exclude-members.test.js
```diff
@@ -903,3 +911,106 @@ test('On exclusion, recover if we fail to re-add anyone at all', async (t) => {
...
+    [aliceId, carolId].sort(),
+    'carol has correct member list'
+  )
+
+  await Promise.all([
+    p(alice.close)(true),
+    p(bob.close)(true),
+    p(carol.close)(true),
+  ])
+})
```
mixmix, line 1016: A beautiful test :peach:

mixmix (approved):
Love it.

test/exclude-members.test.js
mixmix: Feel free to skip this change. It's just a recommended pattern

PR #128, opened 2023-05-30
add commments to case handling

mixmix:


mixmix:
these were comments @Powersource asked for that I forgot to merge.

PR #134, opened 2023-07-17
Be more forgiving in getPreferredEpoch stream

Powersource:
I think this fixes https://github.com/ssbc/ssb-tribes2-demo/issues/7

Manually tested in the demo, seems to make the error go away.

### DIFF
--- a/lib/epochs.js
+++ b/lib/epochs.js
@@ -526,15 +526,13 @@ function getGroupInit(ssb, groupId, cb) {
   pull(
     ssb.box2.getGroupInfoUpdates(groupId),
     pull.take(1),
-    pull.drain((info) => {
-      if (!info) return cb(new Error('Unknown group'))
+    pull.drain(
+      (info) => {
+        if (!info) return cb(new Error('Unknown group'))
 
-      // Fetch the tangle root
-      // Repeating since sometimes the group info comes in very quick, before the root msg has had time to get put into the db
-      getRepeat(
-        ssb,
-        info.root,
-        (err, rootVal) => {
+        // Fetch the tangle root
+        // Repeating since sometimes the group info comes in very quick, before the root msg has had time to get put into the db
+        getRepeat(ssb, info.root, (err, rootVal) => {
           // prettier-ignore
           if (err) return cb(clarify(err, 'Failed to load group root with id ' + info.root))
 
@@ -543,13 +541,13 @@ function getGroupInit(ssb, groupId, cb) {
             return cb(clarify(new Error(isInitRoot.string), 'Malformed group/init root message'))
 
           cb(null, { key: info.root, value: rootVal })
-        },
-        (err) => {
-          // prettier-ignore
-          if (err) return cb(clarify(err, 'Failed to get group info for ' + groupId))
-        }
-      )
-    })
+        })
+      },
+      (err) => {
+        // prettier-ignore
+        if (err) return cb(clarify(err, 'Failed to get group info for ' + groupId))
+      }
+    )
   )
 }
 
--- a/lib/epochs.js
+++ b/lib/epochs.js
@@ -269,10 +269,10 @@ function Epochs(ssb) {
     // This is important for listMembers to not send confusing results
     getPreferredEpoch(groupId, (err, preferredEpoch) => {
       // if we're live and this fails we don't really mind, just go straight to live
-      if (err && !live) return deferredSource.abort(clarify(err, 'failed to get initial preferred epoch'))
-
       if (!live) {
-        deferredSource.resolve(pull.once(preferredEpoch))
+        if (err) deferredSource.abort(clarify(err, 'failed to get initial preferred epoch'))
+        else deferredSource.resolve(pull.once(preferredEpoch))
+
         return
       }
 
--- a/lib/epochs.js
+++ b/lib/epochs.js
@@ -268,14 +268,16 @@ function Epochs(ssb) {
     // then skip all the preferrentEpochs until we get up to the current
     // This is important for listMembers to not send confusing results
     getPreferredEpoch(groupId, (err, preferredEpoch) => {
-      if (err) return deferredSource.abort(clarify(err, 'failed to get initial preferred epoch'))
+      // if we're live and this fails we don't really mind, just go straight to live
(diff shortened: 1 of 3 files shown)


lib/epochs.js
```diff
@@ -509,24 +511,46 @@ function epochNodeStream(ssb, groupId, opts = {}) {
...
-
-      if (!isInitRoot(rootVal))
-        // prettier-ignore
-        return cb(clarify(new Error(isInitRoot.string), 'Malformed group/init root message'))
-
-      cb(null, { key: info.root, value: rootVal })
+  pull(
+    ssb.box2.getGroupInfoUpdates(groupId),
+    pull.take(1),
+    pull.drain((info) => {
```
Powersource, line 529: I suspect this change might not have been needed, but idk if it hurts either?

lib/epochs.js
```diff
@@ -509,24 +511,46 @@ function epochNodeStream(ssb, groupId, opts = {}) {
...
+  ssb.db.get(key, (err, rootVal) => {
+    if (err) {
+      setTimeout(() => {
+        return getRepeat(ssb, key, cb)
+      }, 100)
+    } else {
+      return cb(null, rootVal)
+    }
+  })
+}
```
mixmix, line 524: :fire:  this needs a limit ... like 1 minute, or 100 tries?
100ms might want to be stretched out as well if not in tests?

lib/epochs.js
mixmix: :fire: needs to be able to handle errors on drain

```js
pull.drain(
  (info) => { },
  (err) => { }
)
```

lib/epochs.js
```diff
@@ -509,24 +511,46 @@ function epochNodeStream(ssb, groupId, opts = {}) {
...
+          // prettier-ignore
+          if (err) return cb(clarify(err, 'Failed to load group root with id ' + info.root))
+
+          if (!isInitRoot(rootVal))
+            // prettier-ignore
+            return cb(clarify(new Error(isInitRoot.string), 'Malformed group/init root message'))
+
+          cb(null, { key: info.root, value: rootVal })
+        },
+        (err) => {
```
mixmix, line 547: :fire: this .... is unused. I suspect it was meant to be for the `pull.drain` ??

mixmix (changes requested):
Like this idea. couple of :fire: that need attention then merge

lib/epochs.js
Powersource: oh woops, good catch

lib/epochs.js
Powersource: this is an eventually consistent situation, so might take 1ms or 3hrs. how about exponential backoff?

Powersource: switched to exponential backoff now, with a limit, i think it should give up after a few weeks or something with the current config lol

staltz: The exponential backoff seems like an overkill to solve this race condition (box2 versus db2), and with hours or weeks between "cause" and "effect", this could become a debugging nightmare in the future.

What if you do only a minor exponential backoff (say, with a limit of 1 second, that should be enough to beat the basic db2 race conditions)? And then, what if you only trigger the "add missing members to preferred epoch" logic only if you know you have the group root?

staltz (changes requested):

lib/epochs.js
Powersource: Mm yeah it is a bit weird the long delay. I think I'll just change the backoff to something short even if that could in theory introduce some other problem in the future, but that's for other new tests/reqs/bug reports to figure out.

Powersource: done

Powersource: didn't test this in the demo which i think is where it needs testing, let's see if it needs adjusting though

staltz:
@Powersource If I understood correctly with this exponential-backoff library, the maxTry of 10 means that it'll start with 100ms, then double to 200ms, then 400ms, ..., until it waits 1min and a half. Isn't that too much? Compared to what I was suggesting of 1second maximum, which would be 100ms+200ms+400ms+800ms give or take.

And what about the second mechanism? :

> what if you only trigger the "add missing members to preferred epoch" logic only if you know you have the group root?

mixmix:
Re-running tests.
@Powersource BUMP

Powersource:
reduced to 5 total tries now (including first 0-delay run)

Powersource:
> what if you only trigger the "add missing members to preferred epoch" logic only if you know you have the group root?

isn't that what a long delay here would accomplish? or instead of a long delay we could/should maybe do a stream in that case

staltz:
> isn't that what a long delay here would accomplish? or instead of a long delay we could/should maybe turn this into a stream in that case

Delays are a solution, but a risky one because they add technical debt and debugging puzzles. Once you have one delay as a solution, it's easy to add another, and then the problem compounds.

We should try to fight such kind of technical debt, and whenever there is a data dependency between two async parts of a system, we need to make that dependency explicitly either via streams or promises or things like that. This means two good things: (1) once the first part is ready, the second part will be immediately triggered (this ultimately means that end-user apps are overall quicker to respond), (2) debugging is made simpler since you have a clear cause-and-effect appearing in the stack trace in case of an error. With setTimeouts, the effect is separate from its cause.

Powersource:
so you agree i should turn this into a stream?

Powersource:
should be ready for review again @staltz

lib/epochs.js
```diff
@@ -509,24 +521,66 @@ function epochNodeStream(ssb, groupId, opts = {}) {
...
+    pullMany([
+      ssb.db.query(
+        where(
+          and(
+            isDecrypted('box2'),
+            equal(seekKey, toMessageSigil(msgId), {
+              indexType: 'groupRootMsgId',
+              prefix: true,
+            })
+          )
```
staltz, line 536: `and(key(msgId), isDecrypted('box2')`

staltz (changes requested):

lib/epochs.js
Powersource: done

staltz (approved):
Tests passing too, so, good to merge!

Powersource:
v1.4.0

PR #139, opened 2023-07-25
Disable exclude crash recovery by default

Powersource:
For https://github.com/ssbc/ssb-tribes2/issues/123


staltz (approved):
Yes, looks right

PR #140, opened 2023-09-11
4.6 overlap membership

mixmix:


mixmix:
This is ready for review now @Powersource (all passing)
Recommend we merge and get paid?

lib/epochs/index.js
```diff
@@ -530,46 +432,150 @@ function getGroupInit(ssb, groupId, cb) {
...
+    if (err) return cb(clarify(err, 'Failed to resolve epoch @tangle/reduce'))
+
+    const tips = getTipsFromTangleReduce(reduce)
+    return cb(null, tips)
+  })
+}
+
+function fixDisjointEpochsLater(ssb, groupId) {
+  const timeout =
+    ssb?.config?.tribes2?.disjointResloveDelay ??
```
Powersource, line 507: spelling mistake. and this should be mentioned in the readme

Powersource: and can't this reuse the existing config options we have? https://github.com/ssbc/ssb-tribes2/#config

Powersource: we could also reuse this function https://github.com/ssbc/ssb-tribes2/blob/165967eb2333ff89ed8a6e36e95be1c4b194e6f7/listeners.js#L34-L40

lib/epochs/index.js
```diff
@@ -530,46 +432,150 @@ function getGroupInit(ssb, groupId, cb) {
...
+  const timeoutId = setTimeout(() => {
+    // check if disjoint still
+    getTipEpochsWithMembers(ssb, groupId, (err, tips) => {
+      if (err)
+        return console.error(
+          clarify(err, 'Failed to check disjoint but overlapping state')
+        )
 
-  delete preferredEpoch.members
+      if (!isDisjointOverlappingState(tips)) return
```
Powersource, line 517: don't you already check this on line 469?

Powersource: aah this is a check for after the timeout. but still, can't you use the same logic/function both here and on 469?

Powersource:
review'd

lib/epochs/index.js
mixmix: I'm gonna say "no not mentioning in README" as it's a testing config.
I'm also gonna keep it simple and not reuse that function as this is serving atm

mixmix: Fixed spelling! THanks

lib/epochs/index.js
mixmix: Can't use the logic on line 469 as that is insufficient - note it's part way into an if,else if, else if, so other things have been ruled out earlier.

Keeping this as is for the moment

